PR #10, opened 2022-02-02
Add scatter functionality for loops

SpheMakh:
- Add pathos and dill as deps
-  set `skip_implicits = False` by default for parameters
- Add `for_loop.scatter` functionality


setup.py
```diff
@@ -9,12 +9,18 @@
 requirements = ["pyyaml",
                 "nose>=1.3.7",
                 "future-fstrings",
-                "scabha @ git+https://github.com/caracal-pipeline/scabha2@master#egg=scabha-0.4.0",
+<<<<<<< HEAD
```
o-smirnov: Looks like an unresolved merge.

stimela/__init__.py
```diff
@@ -61,5 +61,5 @@ def register_globals():
 
 from .stimelogging import logger
 
-from stimela.kitchen.recipe import Recipe
+from stimela.kitchen.recipe import PyRecipe as Recipe
```
o-smirnov: Thought we're going to retire PyRecipe?

stimela/commands/exxec.py
```diff
@@ -21,7 +21,7 @@
 
 @cli.command("exec",
     help="""
-    Execute a single cab, or a recipe from a YML file. 
+    Execute a single cab, or a recipe from a YML file_pickle.PicklingError: Can't pickle <class 'types.UpdatedStimelaConfig'>: it's not found as types.UpdatedStimelaConfig.
```
o-smirnov: Ooops where did the error come from? Wild paste?

o-smirnov (changes requested):

stimela/__init__.py
sphemakh: fixed

setup.py
sphemakh: We can put a pypi dep before we release. But this is OK for now since using a release that doesn't exist breaks the install

o-smirnov (approved):

PR #27, opened 2022-03-31
Formula parser & alias cleanup & rich-based help

### LINKED ISSUE #9: implicit outputs  in standard wsclean cab throw "implicit output set explicitly" error message
See same test recipe as #8.

### LINKED ISSUE #17: Add help message for inputs defined in a recipe
Suggested by @landmanbester here https://github.com/ratt-ru/Stimela/issues/745.

It would be very convenient to have the equivalent of a --help for inpits to a stimela recipe. For example
```
$ stimela exec recipe.yaml --help
```
could display the list of available inputs to the recipe.

### LINKED ISSUE #21: Make globs explicit with e.g. a "glob:" prefix
The current scheme for specifying a ``List[File]`` as a glob (see e.g. https://github.com/caracal-pipeline/stimela2/blob/master/stimela/cargo/cab/wsclean.yaml#L21) is very useful, but the internal logic for keeping track of what's a glob and what's an explicit list, and when to expand the glob, is getting a little twisted. It would be easier if we had an explicit way to tell that something needs to be treated as a glob. For example, the relevant part of wsclean's schema would look like this:

```yml
  dirty:
    info: Dirty images
    dtype: List[File]
    implicit: "glob:{current.prefix}-[0-9][0-9][0-9][0-9]-dirty.fits"
    must_exist: false
  restored: 
    info: Restored images
    dtype: List[File]
    implicit: "glob:{current.prefix}-[0-9][0-9][0-9][0-9]-image.fits"
    must_exist: false
```

@SpheMakh what do you think? Going forward, a similar convention could be applied to specify expressions that need to be evaluated, e.g.

```yml
  x:
    implicit: 'eval:recipe.x+current.z'
```

### LINKED ISSUE #23: support for *.name aliases
@landmanbester has a use case in his PFB recipes where he wants a common parameter (``nthreads``) defined at recipe level and passed to many steps at once.

It would be fairly simple to allow something like ``*.nthreads`` as an alias target. I.e. an alias to multiple steps matching the glob.

### LINKED ISSUE #25: Optional skip_if_exists attribute for a step?
I'm using the start step/end step feature quite a lot, because often I don't want to repeat e.g. initial imaging steps.

It occurs to me that this attribute of a step would be useful:

```yml
skip_if_exists: [foo, bar]
```

The logic then would be quite simple -- before running a step, check if the outputs referred to by ``foo`` and ``bar`` already exist (obviously, this only applies to file/Dir/MS outputs), and skip the step if they do.

@SpheMakh @landmanbester thoughts?

o-smirnov:
Matched set with https://github.com/caracal-pipeline/scabha2/pull/21.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -29,7 +29,7 @@ setup(name=PACKAGE_NAME,
       url="https://github.com/ratt-ru/Stimela",
       packages=find_packages(),
       include_package_data=True,
-      python_requires='>=3.6',
+      python_requires='>=3.7',
       install_requires=requirements,
       entry_points="""
             [console_scripts]
--- a/setup.py
+++ b/setup.py
@@ -16,6 +16,7 @@ requirements = ["pyyaml",
                 "click",
                 "pydantic",
                 "pathos",
+                "psutil"
                 ],
 
 PACKAGE_NAME = "stimela"
--- a/setup.py
+++ b/setup.py
@@ -16,7 +16,8 @@ requirements = ["pyyaml",
                 "click",
                 "pydantic",
                 "pathos",
-                "psutil"
+                "psutil",
+                "rich"
                 ],
 
 PACKAGE_NAME = "stimela"
--- a/setup.py
+++ b/setup.py
@@ -9,7 +9,7 @@ import glob
 requirements = ["pyyaml",
                 "nose>=1.3.7",
                 "future-fstrings",
-                "scabha @ git+https://github.com/caracal-pipeline/scabha2",
+                "scabha >= 0.6.1", # @ git+https://github.com/caracal-pipeline/scabha2",
                 "ruamel.yaml",
                 "munch",
                 "omegaconf>=2.1pre1",
--- a/setup.py
+++ b/setup.py
@@ -29,7 +29,7 @@ setup(name=PACKAGE_NAME,
       url="https://github.com/ratt-ru/Stimela",
       packages=find_packages(),
       include_package_data=True,
-      python_requires='>=3.7',
+      python_requires='>=3.6',
       install_requires=requirements,
       entry_points="""
             [console_scripts]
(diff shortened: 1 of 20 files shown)


setup.py
```diff
@@ -9,13 +9,19 @@
 requirements = ["pyyaml",
                 "nose>=1.3.7",
                 "future-fstrings",
-                "scabha @ git+https://github.com/caracal-pipeline/scabha2",
+                "scabha >= 0.7.0",
```
sphemakh, line 12: Then we should just remove the scabha dependency for now. Because scabha 0.7.0 is not on pypi, and it makes no sense to require it.

stimela/commands/run.py
```diff
@@ -1,22 +1,25 @@
...
 from yaml.error import YAMLError
 from scabha import configuratt
 from scabha.exceptions import ScabhaBaseException
 from omegaconf.omegaconf import OmegaConf, OmegaConfBaseException
+from stimela import stimelogging
 from stimela.config import ConfigExceptionTypes
 import click
 import logging
-import os.path, yaml, sys
+import os.path, yaml, sys, asyncio
```
sphemakh: Not kosher. Unless importing from the same module/library, each import should be on a new line

SpheMakh (changes requested):
I like the new features,  especially `xrun_asyncio`!

Let's remove (comment out) the scabha dependency from setup.py for now.

setup.py
o-smirnov: Or we could just merge the scabha PR and make a release?

stimela/commands/run.py
o-smirnov: PEP-8 says "Imports should *usually* be on separate lines", so I tend to interpret that loosely. Particularly when a large number of the usual standard library suspects are imported, like ``import os, os.path, sys`` etc. I just find the import section gets too big.

o-smirnov:
@SpheMakh please approve and merge.

setup.py
sphemakh: That works for me

PR #43, opened 2022-06-14
Can nest sub-recipes simply by name now. And prettier error reporting.

o-smirnov:

PR #45, opened 2022-06-15
Cache configs and recipes for faster startup

o-smirnov:
Uses the configuratt cache (https://github.com/caracal-pipeline/scabha2/pull/31) to make startup really fast.

This includes and supersedes #43.


o-smirnov:
Superceded by #48.

PR #72, opened 2022-10-24
Setup readthedocs

SpheMakh:


docs/source/conf.py
```diff
@@ -0,0 +1,58 @@
...
+
+# -- General configuration ---------------------------------------------------
+# https://www.sphinx-doc.org/en/master/usage/configuration.html#general-configuration
+
+extensions = []
+
+templates_path = ['_templates']
+exclude_patterns = []
+
+language = 'ls -lrt'
```
sjperkins: Maybe should be English?

docs/source/conf.py
```diff
@@ -0,0 +1,58 @@
...
+# https://www.sphinx-doc.org/en/master/usage/configuration.html
+
+# -- Project information -----------------------------------------------------
+# https://www.sphinx-doc.org/en/master/usage/configuration.html#project-information
+
+
+import stimela
+
+project = 'stimela'
+copyright = '2022, Sphesihle Makhathini, Oleg Smirnov and RATT'
```
sjperkins: If it's convenient, you could import datetime and programmatically generate the year  in an f-string.

readthedocs.yaml
```diff
@@ -0,0 +1,20 @@
...
+
+# Set the version of Python and other tools you might need
+build:
+  os: ubuntu-20.04
+  tools: {python: "3.8"}
+  jobs:
+    post_install:
+      - pip install poetry>=1.2.1
+      - poetry config virtualenvs.create false
+      - poetry install --with docs
```
sjperkins, line 15: OK this is going to depend on a poetry group called `docs` which isn't present in `pyproject.toml` . You can also read here for more info: https://python-poetry.org/docs/managing-dependencies/#dependency-groups and probably dask-ms's `pyproject.toml` for how to create it.

Groups don't have a `setup.py` correspondence, they're a new poetry feature. There's potential for confusion here.

Note that poetry also has the concept of [extras](https://python-poetry.org/docs/pyproject#extras) which *does* correspond to `setup.py`'s `extras_require` feature. I see you've used it for the `pytest` dependency.

docs/source/conf.py
```diff
@@ -0,0 +1,58 @@
...
+
+templates_path = ['_templates']
+exclude_patterns = []
+
+language = 'ls -lrt'
+
+# -- Options for HTML output -------------------------------------------------
+# https://www.sphinx-doc.org/en/master/usage/configuration.html#options-for-html-output
+
+html_theme = 'alabaster'
```
sjperkins: [furo](https://github.com/pradyunsg/furo) is a nice theme but you'll have to add it to the `docs` group mentioned elsewhere in this review.

```suggestion
html_theme = 'furo'
```

I don't think this is critical though.

docs/source/conf.py
```diff
@@ -0,0 +1,58 @@
...
+# extensions coming with Sphinx (named 'sphinx.ext.*') or your custom ones.
+extensions = [
+    "sphinx_copybutton",
+    "sphinx.ext.autodoc",
+    "sphinx.ext.autosummary",
+    "sphinx.ext.doctest",
+    "sphinx.ext.intersphinx",
+    "sphinx.ext.extlinks",
+    "numpydoc",
+]
```
sjperkins, line 49: One thing to watch out for is any packages with C-extensions (numpy is often a culprit here).  https://docs.readthedocs.io/en/stable/faq.html#i-get-import-errors-on-libraries-that-depend-on-c-modules.

In that case you'll probably have to hunt the offending imports down and then place them in something like this:

```python
autodoc_mock_imports = ["pyrap", "s3fs", "xarray"]
```

https://github.com/ratt-ru/dask-ms/blob/53ab607b5ef727d6366b115f8f3060bf61efd603/docs/conf.py#L28

sjperkins (changes requested):
A poetry docs group will need to be added for RTD to begin working, otherwise I've got some low priority suggestions.

I haven't reviewed the Python or cab changes

sphemakh:
Ready for final review

PR #74, opened 2022-10-26
Kube rekube

o-smirnov:
@SpheMakh wanna do the honours?


o-smirnov:
Actually hang on, seeing one more glitch with for-loops, gimme a few minutes...

PR #80, opened 2022-10-31
Scatter, redone

o-smirnov:
@landmanbester you'll want to try this one too.

I've gone and fixed the scatter mechanism. Changed to using https://docs.python.org/3/library/concurrent.futures.html, as this allows nested process pools, and we can drop the pathos dependency.

I've also changed the definition of scatter to be an int, specifying the max number of workers. If scatter=-1, all for-loop iterations are scattered.

Finally, ``over`` and ``scatter`` will also be looked up within recipe inputs. So the following two are equivalent:

```yml
recipe_name:
  for_loop:
    var: x
    over: [1,2,3]
    scatter: -1
```

and

```yml
recipe_name:
  for_loop:
    var: x
  inputs:
    for_loop:
      over: 
        dtype: List[int]
        default: [1,2,3]
      scatter: 
        dtype: int
        default: -1
```
...but the latter construct allows the list of iterations (and the scatter parameter) to be passed to the recipe at run-time (i.e. from the command line, or from a previous step of a recipe, if the for-loop is in a sub-recipe). This should help things like https://github.com/caracal-pipeline/stimela2/issues/77

### DIFF
--- a/stimela/kitchen/recipe.py
+++ b/stimela/kitchen/recipe.py
@@ -34,9 +34,10 @@ class ForLoopClause(object):
     # name of list variable
     var: str 
     # This should be the name of an input that provides a list, or a list
-    over: Any
-    # If True, this is a scatter not a loop -- things may be evaluated in parallel
-    scatter: bool = False
+    over: Optional[Any] = None
+    # If !=0 , this is a scatter not a loop -- things may be evaluated in parallel using this many workers
+    # (use -1 to scatter to unlimited number of workers)
+    scatter: int = 0
 
 def IterantPlaceholder(name: str):
     return name
@@ -120,6 +121,8 @@ class Recipe(Cargo):
         # set of keys protected from assignment
         self._protected_from_assign = set()
         self._for_loop_values = None
+        # process pool used to run for-loops
+        self._loop_pool = None
 
     def protect_from_assignments(self, keys):
         self._protected_from_assign.update(keys)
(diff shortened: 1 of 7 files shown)


stimela/kitchen/recipe.py
```diff
@@ -1,27 +1,25 @@
-from cmath import exp
-from multiprocessing import cpu_count
-import os, os.path, re, logging, fnmatch, copy, time
+import os, os.path, re, fnmatch, copy, traceback
```
sphemakh, line 1: SAD

stimela/kitchen/recipe.py
```diff
@@ -1,27 +1,25 @@
...
 
-from pathos.pools import ProcessPool
-from pathos.serial import SerialPool
-from multiprocessing import cpu_count
+from concurrent.futures import ProcessPoolExecutor
+
 from stimela.config import EmptyDictDefault, EmptyListDefault
 import stimela
 from stimela import log_exception, stimelogging
 from stimela.exceptions import *
```
sphemakh, line 14: SAD

SpheMakh (approved):

landmanbester:
How do I specify distributive details? So, for example, I want to calibrate ms1 on node 1 and ms2 on node 2, how do I configure the number of threads etc. on the different nodes?

stimela/kitchen/recipe.py
o-smirnov: ["The Ten Commandments. She was a great believer in some of them."](https://libquotes.com/joe-orton/quote/lbv2c5y)

That's me and PEP-8!

o-smirnov:
@landmanbester that's still TBD. Depends on what state the slurm runner is in (@SpheMakh?) We'd need to design in some optional attributes to determine that. It's not difficult to do that, but let's brainstorm a little to converge to a neat solution.

At the moment all this does is run each step of the for-loop in parallel. In the native case, this means each binary is spawned in parallel. In the Kubernetes case, it means pods will be instantiated simultaneously.

Issue #82, opened 2022-11-01
inline Python code in cab command


o-smirnov:
As suggested by @SpheMakh at the retreat, this would be useful:

```yaml
cab:
  flavour: python-code
  command: |
    from casacore.tables import table
    tab = table("{current.ms}")
    tab.do_funny_stuff()
  inputs:
    ms: 
      dtype: MS
      required: true
```

Bit of an open question how such code would return outputs. Related to #78.

o-smirnov:
Also related to #20 and #46. Thinking of four flavours now, but we need to take some decisions on names:

* ``python``, what we call ``python-ext`` right now: call Python function specified as ``command: module_name.function_name`` in the cab definition

* ``python-code``: call inlined block of Python as per above

* ``casa-task``: call specific CASA task, with inputs explicitly defined by the cab

* ``casa-code``: call casapy and have it run an inlined block of Python

Issue #78, opened 2022-10-29
add support for rich return values from cabs


o-smirnov:
Possibilities:

* Extend output wranglers to parse values out of the console output with regexes

* Let a cab return a YaML mapping of outputs

* Let a Python cab return a dict

o-smirnov:
> * Let a Python cab return a dict

This would only work with the inline Python flavour (see https://github.com/caracal-pipeline/stimela2/issues/20) and won't work with non-native runners. So not a good way at all.

> * Let a cab return a YaML mapping of outputs

Noting that the only route of communication from an external process is its stdout, we're anyway going to have to

> * Extend output wranglers to parse values out of the console output with regexes

This is straightforward to implement, but the cab must print stuff in a specifically formatted way. So, for example:

```python
print(f"CAB_OUTPUT: {output_name}: {json.dumps(value)}"
```

...and I could then write a standard wrangler that would parse this out into an output. Perhaps a better look for this would be (in the cab):

```python
from stimela.scabha.cab_utils import yield_output
yield_output(name, value)
```

PR #88, opened 2022-11-03
Wrangler update, can parse outputs

o-smirnov:
This implements #78, and opens the way to #82.

This also obviates ``crasa``, since everything can now be done with a few standard wrangler definitions within CASA cabs.


o-smirnov:
This is also included in #89, so @SpheMakh you can rather review that so long.

SpheMakh (approved):

PR #91, opened 2022-11-05
Cleaned up & simplified & prettified logging machinery

### LINKED ISSUE #78: add support for rich return values from cabs

o-smirnov:


SpheMakh (approved):
Is the option to delete an output product before running a cab part of this PR, or are we adding that later?

o-smirnov:
> Is the option to delete an output product before running a cab part of this PR, or are we adding that later?

No, good point, it isn't -- let's leave https://github.com/caracal-pipeline/stimela2/issues/46 open so long until I fix this.

PR #93, opened 2022-11-08
allow key=value command-line options

SpheMakh:


stimela/kitchen/cab.py
```diff
@@ -243,7 +243,8 @@ def stringify_argument(name, value, schema, option=None):
                 # check repeat policy and form up representation
                 repeat_policy = get_policy(schema, 'repeat')
                 if repeat_policy == "list":
-                    return [option] + list(value) if option else list(value)
+                    value = [" ".join(value)]
```
o-smirnov, line 246: @SpheMakh I think this change will break tools that expect --option X Y, since in this case three separate arguments must be passed. We're not using shell argument parsing, so it won't break up "X Y" into two arguments.

stimela/kitchen/cab.py
```diff
@@ -313,12 +315,19 @@ def stringify_argument(name, value, schema, option=None):
...
                 else:
                     explicit = get_policy(schema, 'explicit_true' if value else 'explicit_false')
                     args += [option, str(explicit)] if explicit is not None else ([option] if value else [])
             else:
+                
                 value = stringify_argument(name, value, schema, option=option)
                 if type(value) is list:
+                    if key_value_command_line:
+                        if len(value) > 1:
+                            value = [f"{value[0]}={value[1]}"]
```
o-smirnov, line 330: What if ``value`` has more than two elements? This will result in everything past the second element being ignored. Probably better to throw an error here, since for an input of [X,Y], ``--option=X Y`` is not a sensible construction anyway. More generally, ``repeat_policy=list`` can't really work with key=value syntax in any sensible way, so we should flag that up front during cab validation.

stimela/kitchen/cab.py
```diff
@@ -313,12 +315,19 @@ def stringify_argument(name, value, schema, option=None):
             if schema.dtype == "bool":
                 if key_value:
                     args += [f"{name}={value}"]
```
o-smirnov, line 317: Spotted an old bug here -- ``{name}`` should be ``{schema.nom_de_guerre or name}``.

o-smirnov (changes requested):
@SpheMakh, is the point of this to support ``--option=value`` constructs? Rather than introducing an extra policies property, should we not rather have ``prefix='--'``, ``key_value: true`` to represent this case (and empty ``prefix``, ``key_value: true`` for the ``option=value`` case)?

stimela/kitchen/cab.py
sphemakh: > should we not rather have key_value: true to represent this case (and empty prefix, key_value: true for the option=value case)? 

That's what I tried at first, but it doesn't work since `key_val` policy does not treat lists and `explicit_{true|false}` cases. I'm assuming `key_val` was not really meant for a command-line parser.

stimela/kitchen/cab.py
sphemakh: Yes, this list can only have two elements in this case. Could raise an exception if this is not the case.

stimela/kitchen/cab.py
sphemakh: ahh, didn't think about that. Will fix

stimela/kitchen/cab.py
o-smirnov: > I'm assuming key_val was not really meant for a command-line parser.

But it was -- there are tools (e.g. QuartiCal) that use key=value command lines. So any problems should be treated as bug to be fixed. ``explicit_{true|false}`` was meant to work with it. As for lists, ``repeat_policy=list`` makes no sense with this construct as far as I can see (see other remark in this review), but other repeat_policies should work.

sphemakh: Noted, I'll add expand the functionality of `key_val`. Making this a draft PR for now, will re-open later

PR #120, opened 2023-01-02
What If I want to use a commandline debugger for dev?

Mulan-94:
- Adds dev option to switch animated progress bar off. To be accessed by:
`stimela --dev run blabla`
not sure if this is the way to do it, but works well for me
- Small fixe for some string parsing in casa flavour backend
- Small error message fix

### DIFF
--- a/stimela/main.py
+++ b/stimela/main.py
@@ -53,10 +53,11 @@ class RunExecGroup(click.Group):
 @click.option('--clear-cache', '-C', is_flag=True, 
                 help="Reset the configuration cache. First thing to try in case of strange configuration errors.")
 @click.option('--verbose', '-v', is_flag=True, help='Be extra verbose in output.')
+@click.option('--dev', '-dev', is_flag=True, help='Dev mode, disables progress bar allowing cmd line debug.')
 @click.version_option(str(stimela.__version__))
-def cli(backend, config_files=[], config_dotlist=[], include=[], verbose=False, no_sys_config=False, clear_cache=False):
+def cli(backend, config_files=[], config_dotlist=[], include=[], verbose=False, no_sys_config=False, clear_cache=False, dev=False):
     global log
-    log = stimela.logger(loglevel=logging.DEBUG if verbose else logging.INFO)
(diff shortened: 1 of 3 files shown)


stimela/commands/run.py
```diff
@@ -166,7 +166,7 @@ def run(what: str, parameters: List[str] = [], dry_run: bool = False, profile: b
                 try:
                     stimela.CONFIG.lib.recipes[name] = OmegaConf.merge(RecipeSchema, value)
                 except Exception as exc:
-                    log_exception(f"error in definition of recipe '{name}'", exc)
+                    log_exception(f"error in definition of recipe '{value}'", exc)
```
o-smirnov: @Mulan-94 won't this dump the entire recipe into the exception string? Doesn't that make it super verbose and unreadable?

o-smirnov (commented):
Unsure about the verbose error message, but happy with the other changes!

stimela/commands/run.py
Mulan-94: I'm not really sure. I changed it to the current state because I was getting an error message like:

`error in definition of recipe 'name'`

for an error in my recipe, which was not very clear. I assumed the idea was to pass whatever is held by the  'name' key. But, I haven't tested  how much can be dumped, shall I change it back?

o-smirnov: Well it should have been reporting the name of the top level recipe section, not just literally ``'name'``. Unless your top-level section was actually called "name"?

sphemakh:
@Mulan-94 can you adress the merge conflicts in this PR

sphemakh:
@Mulan-94, your PR is getting stale. Are you still working on this?

Mulan-94:
Oh yeah, i forgot about this, it was complete. Let me update with master

Mulan-94:
@SpheMakh I just did a push with the current master and included the dev mode only

landmanbester:
This seems very useful can we merge it please?

o-smirnov:
Equivalent fix committed onto ``FIASCO3`` branch.

Issue #127, opened 2023-01-12
allow a "gentle" Ctrl+C interrupt


o-smirnov:
When running a cab, the first Ctrl+C shouldn't stop it outright -- rather, let the task finish, and mark the pipeline to be interrupted at the end of it.

A second Ctrl+C should stop it. This is easy to indicate in the status line.

o-smirnov:
I still like it!

gijzelaerr:
**Sprint triage: death row**. 3+ years old, no traction. Graceful interrupt handling is valid but this issue has no concrete design or implementation path. Can be re-filed if someone wants to tackle it. Recommend closing.

o-smirnov:
I think this is still useful and I'm curious what Claude can make of this without further prompting, let's add it to the sprint.

PR #131, opened 2023-01-12
fixes #126. Fixes #129. Fixes #130. Fixes #125. Fixed #116. 

### LINKED ISSUE #116: add option to skip a step if all outputs already exist
Lengthy one-off steps that generate files could be specified as skipped automatically if the outputs already exist, this is an easy check. I've already been doing this by hand in my recipes via e.g.

```yml
skip: =EXISTS(current.output_name)
```

But this could be made simpler by recognizing e.g.:

```yml
skip: =OUTPUTS_EXIST
```

Another option could be:

```yml
skip: =OUTPUTS_FRESH
```

...which would skip the step if outputs exist, and are all newer than the inputs (old school Makefile-style).

### LINKED ISSUE #125: config caching mechanism inteferes when top-level recipes _include other top-level recipes
Solution is to not load the top-level recipe from cache in stimela run.

### LINKED ISSUE #126: use of UNSET and unresolved substitutions lead to inconsistency in for-loops
Solution incoming:

* more consistent treatment of ``Unresolved`` objects, do not attempt substitutions on them

* rather than using ``UNSET`` to unset parameters, use ``=UNSET`` (i.e. invoke the formula parser). The word ``UNSET`` should not have a special meaning outside of a formula.

### LINKED ISSUE #129: better status display for for-loops
Current display status when running a for loop is ``[N/M]``.

Looping over scans in my Jove reduction, I like to see the scan name too, so e.g. ``[scan=11 2/4]``.

This can be supported by an optional ``display_status`` field under ``for_loop``. If not given, use ``[N/M]``. Otherwise, specify a {}-format string, so e.g. the above can be accomplished by

```yml
  for_loop:
    var: scan
    over: scan-list
    display_status: "{var}={value} {index1}/{total}"
```

### LINKED ISSUE #130: allow PARSE_OUTPUT wrangler to specify regex group name explicitly
PARSE_OUTPUT would be easier to use if one could (optionally) specify a named regex group (or group number) separately from the output name.

o-smirnov:
Initial steps towards caracal-pipeline/stimela#127 and caracal-pipeline/cult-cargo#1, too.

### DIFF
--- a/stimela/commands/run.py
+++ b/stimela/commands/run.py
@@ -13,6 +13,7 @@ from omegaconf.omegaconf import OmegaConf, OmegaConfBaseException
 
 import stimela
 from scabha import configuratt
+from scabha.basetypes import UNSET
 from scabha.exceptions import ScabhaBaseException
 from stimela import stimelogging
 import stimela.config
@@ -90,11 +91,17 @@ def run(what: str, parameters: List[str] = [], dry_run: bool = False, profile: O
     errcode = 0
     recipe_name = None
 
+    def convert_value(value):
+        if value == "=UNSET":
+            return UNSET
+        else:
+            return yaml.safe_load(value)
+
     # parse assign values as YaML
     for key, value in assign:
         # parse string as yaml value
         try:
-            params[key] = yaml.safe_load(value)
+            params[key] = convert_value(value)
         except Exception as exc:
             log_exception(f"error parsing value for --assign {key} {value}", exc)
             errcode = 2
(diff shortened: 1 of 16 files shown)


stimela/commands/run.py
```diff
@@ -362,5 +374,6 @@ def elapsed():
     task_stats.save_profiling_stats(outer_step.log, 
             print_depth=profile if profile is not None else stimela.CONFIG.opts.profile.print_depth,
             unroll_loops=stimela.CONFIG.opts.profile.unroll_loops)
-
+    
+    outer_step.log.info(f"last log directory was [bold green]{stimelogging.get_logfile_dir(outer_step.log) or '.'}[/bold green]")
     return 0
```
sphemakh: Please add newline

SpheMakh (changes requested):

stimela/commands/run.py
o-smirnov: Done.

SpheMakh (approved):

PR #172, opened 2023-08-30
FIASCO III: Kubernetes backend functional

### LINKED ISSUE #158: Parameter validation when using clickify_parameters for CL interfaces
Ran into a limitation of clickify_parameters for constructing CLI's in pfb-clean viz. it doesn't do any parameter validation. So, suppose I add the following to the inputs of a cab definition
```
inputs:
  param:
    dtype: List[int]
```
that I want to run as a command line program. I can't in this case specify a list of ints like
```
$ program --param [0,10]
```
it is currently interpreted as a string. I can wrangle this into a list myself but that entails different behaviour depending on whether I run the program from the terminal or through stimela. It looks like click supports some version of parameter pre-validation

https://click.palletsprojects.com/en/8.1.x/options/#callbacks-for-validation

so this could be a cool feature to add

o-smirnov:


o-smirnov:
@SpheMakh I don't think you're keeping up enough to possibly meaningfully review this... shall I just use superpowers to do a bulk merge to master? This branch has effectively been our master for a long time anyway.

sphemakh:
Sure, please merge. But let's not do these bulky PRs moving forward.

SpheMakh (approved):

o-smirnov:
> Sure, please merge. But let's not do these bulky PRs moving forward.

Only if you promise to review as fast as we code! :rofl: But also @Athanaseus can help next time.

I've pushed an 2.0rc8 release, and will update cult-cargo accordingly.

PR #182, opened 2023-09-21
Propagate imagePullPolicy for worker pods

landmanbester:
This just ensures that imagePullPolicy is propagated also for worker pods. Without it you can end up with stale images lingering on instances where they are already present. Mainly useful during dev


stimela/backends/kube/daskjob.py
```diff
@@ -177,14 +177,14 @@ def daskjob_template(args):
                                         "$(DASK_WORKER_NAME)",
                                         "--nworkers", '1',
                                         "--nthreads", str(args.threads_per_worker),
-                                        "--memory-limit", str(args.memory_limit),
+                                        "--memory-limit", str(0),  #str(args.memory_limit),
```
o-smirnov, line 180: Why's this?

landmanbester: Oh sorry, forgot about that. Dask will start spilling to disk at the stupidly configured 60% default without it. I guess we need a way to to set this in kubeconfig instead of hardcoding it. Actually it would be great if there was a way to expose more of the cluster config in kubeconfig

sphemakh:
@landmanbester is this PR still valid?

landmanbester:
Well I see that the imagePullPolicy is still hardcoded to IfNotPresent for worker pods. Has anyone come up with a more elegant solution to the original problem? If so please close, otherwise I will update the branch

o-smirnov:
It should be a minor fix, I'll take a look shortly.

o-smirnov:
Superseded by #205.

PR #201, opened 2024-01-26
Minor changes for compatibility with Python 3.11.

JSKenyon:
As it says on the label. The only problem is that I am not sure whether or not there will be unforeseen consequences to these changes. I had to make some subtle changes in QaurtiCal to accommodate these fixes. @o-smirnov I would appreciate your input.


o-smirnov (approved):
Yep, this is pretty standard.

o-smirnov:
Thanks @JSKenyon!

PR #204, opened 2024-01-29
Issue 203

o-smirnov:
Redoes #182 properly.


o-smirnov:
Grr nvm, let me base it off master rather.

PR #228, opened 2024-02-11
Issue 198

### LINKED ISSUE #198: _include should support explicit "include from CWD" construct
As picked up by @landmanbester:

Let's say package ``foo`` contains ``X.yml`` and ``Y.yml``, and ``X.yml`` does ``_include: Y.yml``.

Now we have a ``./test.yml``recipe containing ``_include: (foo)X.yml``. X is included correctly, but the results of its nested include of Y will depend on whether ``.`` contains a local ``Y.yml`` of its own or not -- because as implemented currently, the ``_include`` statement looks in the CWD first, then in a standard set of include paths, and then only in the package directory. So it will preferentially pick up a "local" version of Y instead of the package version. This can lead to confusing results.

On the other hand, there is a legit use case for "import local Y.yml if it exists, else use the packaged Y.yml".

Python solves the same problem with package/module paths by supporting an ``import .Z`` or ``from . import Z`` construct, the ``.`` indicating "I really do mean to import from here and not anywhere else!" 

I suppose a Pythonic solution would be to support an explicit ``_include: (.)Y.yml`` construct with the same semantics. This would force it to import Y from the directory of the containing file (and throw an error if it's not found), while ``_include: Y.yml`` would retain the current behaviour of looking in the standard set of include paths, starting from the CWD.

@landmanbester opinions?

o-smirnov:
Fixes #198. Including a check for recursive inclusion.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "stimela"
-version = "2.0rc9"
+version = "2.0rc10"
 description = "Framework for system agnostic pipelines for (not just) radio interferometry"
 authors = ["Sphesihle Makhathini <sphemakh@gmail.com>", "Oleg Smirnov and RATT <osmirnov@gmail.com>"]
 readme = "README.rst"
--- a/scabha/configuratt/__init__.py
+++ b/scabha/configuratt/__init__.py
@@ -18,6 +18,7 @@ def load(path: str, use_sources: Optional[List[DictConfig]] = [], name: Optional
         includes: bool=True, selfrefs: bool=True, include_path: str=None, 
         use_cache: bool = True,
         no_toplevel_cache = False,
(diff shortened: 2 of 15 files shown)


o-smirnov:
Please review #227 first.

SpheMakh (dismissed):

o-smirnov:
"The merge-base changed after approval", yeah, cascading PRs are creating a bit of a pain I'm finding. I'll try to modify my workflow to avoid them in the future.

Athanaseus (dismissed):

o-smirnov:
OK it's the same

> "The merge-base changed after approval", yeah, cascading PRs are creating a bit of a pain I'm finding. I'll try to modify my workflow to avoid them in the future.

All right, I'm going to assume successful reviews and force-merge the lot.

PR #244, opened 2024-02-21
Userutil

### LINKED ISSUE #222: Add module to ease the *scabha* barrier to entry
I'm using scabha as a parser for simms 2.0 and added [this module](https://github.com/wits-cfa/simms/blob/2c8baf23f89504c7f280b16252df5064df5957b3/simms/parser_config/utils.py) to help load parameter specification files

A runner then uses it to get a command-line parser as follows:

````python
import simms
from simms.parser_config.utils import load, load_sources
from scabha.schema_utils import clickify_parameters
import click
from omegaconf import OmegaConf

command = "observe"
sources = load_sources(["skysim", "telescope"])
config = load(command, use_sources=sources)


@click.command(command)
@click.version_option(str(simms.__version__))
@clickify_parameters(config)
def runit(**kwargs):
    opts = OmegaConf.create(kwargs)
    print(opts)
    
runit()
````

I think my `utils.py` module (or some version of it) linked here could be added to *scabha* to make things easier for casual users who want to adopt it as an argument parser.

### PR DESCRIPTION
**Pull request by SpheMakh**
Fixes #222 

For an example, [see this](https://github.com/SpheMakh/contsub/blob/dce81ea1685781034ad88563dfe4def3a5c20992/contsub/parser/imcontsub.py#L9-L21)

### DIFF
--- a/scabha/schema_utils.py
+++ b/scabha/schema_utils.py
@@ -4,10 +4,11 @@ import click
 from scabha.exceptions import SchemaError
 from .cargo import Parameter, UNSET, _UNSET_DEFAULT
 from .basetypes import EmptyDictDefault, File
-from typing import List, File, Dict, DefaultDict, Any
+from typing import List, Union, Optional, Callable, Dict, DefaultDict, Any
 from dataclasses import dataclass, make_dataclass, field
 from omegaconf import OmegaConf, MISSING
 from collections.abc import MutableSet, MutableSequence, MutableMapping
+from scabha import configuratt
 
 def schema_to_dataclass(io: Dict[str, Parameter], class_name: str, bases=(), post_init: Optional[Callable] =None):
     """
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: scabha/schema_utils.py [hunk 1]
```diff
@@ -210,3 +213,39 @@ def clickify_parameters(schemas: Union[str, Dict[str, Any]]):
...
 
     return decorator_chain
+
+@dataclass
+class SchemaSpec:
+    inputs: Dict[str, Parameter]
+    outputs: Dict[str, Parameter]
+    libs: Dict[str, Any]
+
+def paramfile_loader(paramfiles: List[File], sources: List[File] = [], schema_spec=None, use_cache=False) -> Dict:
```
o-smirnov: Suggest Union[File|List[File]], since that's how you treat it in the code below. Just to make the function signature more self-explanatory.

o-smirnov (dismissed):

o-smirnov:
Happy to merge, the above suggestion is purely cosmetic.

o-smirnov (approved):
LGTM

o-smirnov:
@SpheMakh ooops that broke things. And I missed it in my review. I blame MK@5 :rofl:

o-smirnov:
``File|List[File]`` should be ``Union[File, List[File]]``.

sphemakh:
Ahh, the `|` operator is the same as a `Union`, so `int|float` is the same as `Union[int|float]`. But I jumped the gun because this is only implemented in versions 3.10 and above. 

But how did the packaging github action miss this 🤔

o-smirnov:
I'm not sure.  Does it run a linter?

o-smirnov:
I think it's because none of the current tests import ``schema_utils``. The error is thrown upon import. Anyway, easy fix, please approve #248.

PR #256, opened 2024-03-10
Big documentation push

o-smirnov:
Also a couple of tiny code fixes.

After review, let us release this as 2.0rc15 so that readthedocs gets updated. Then one final read and we go for R2.0?


SpheMakh (commented):
All good

SpheMakh (approved):

PR #266, opened 2024-03-26
Add recursive finder for filelike objects

JSKenyon:
(Hopefully) does what it says on the tin. As @o-smirnov points out, this may be somewhat error prone. That said, I have added a handful of tests and the pipeline which originally identified the problem now runs through. In principle, the code could be further refined (the `typeguard` package seems to be very useful), but I think that this should be enough for now. If more exotic types foil this approach, we can always add them to the test suite and iterate.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -24,6 +24,7 @@ pydantic = "^1.10.2"
 psutil = "^5.9.3"
 rich = "^13.7.0"
 dill = "^0.3.6"
+typeguard = "^4.2.1"
 
 [tool.poetry.scripts]
 stimela = "stimela.main:cli"
--- a/scabha/basetypes.py
+++ b/scabha/basetypes.py
@@ -1,11 +1,15 @@
+from __future__ import annotations
 from dataclasses import field, dataclass
 from collections import OrderedDict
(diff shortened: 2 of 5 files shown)


JSKenyon:
Requires #268.

JSKenyon:
@o-smirnov Tests are now passing (rebased on master after dropping 3.7). Feel free to add/change at your leisure.

o-smirnov:
OK, I also added the following option to allow explicit user-specified binding of directories. Simple enough change.

```yml
opts:
  backend:
    singularity:
      bind_dirs:
        ~/tmp: rw
        /tmp: ro 
```

@JSKenyon if the pipeline is in front of you now, could you please test this quickly?

JSKenyon:
> @JSKenyon if the pipeline is in front of you now, could you please test this quickly?

In progress. Have disabled my file finder and I am instead mounting everything manually. Will let you know once it runs through.

JSKenyon:
@o-smirnov That seems to have worked. The combination of the finder and the override should be pretty foolproof.

o-smirnov (approved):
LGTM

Athanaseus (approved):
Nice!

PR #267, opened 2024-03-27
Escape messages before emitting.

JSKenyon:
This is a potential (but possibly imperfect) fix for log messages containing characters which could be misinterpreted as markup tags. This issue was uncovered when specifying a list of absolute paths. Specifically, `[\path]` will be misinterpreted as a markup tag and cause `stimela` to crash.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "stimela"
-version = "2.0rc17"
+version = "2.0rc18"
 description = "Framework for system agnostic pipelines for (not just) radio interferometry"
 authors = ["Sphesihle Makhathini <sphemakh@gmail.com>", "Oleg Smirnov and RATT <osmirnov@gmail.com>"]
 readme = "README.rst"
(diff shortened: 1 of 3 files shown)


JSKenyon:
@o-smirnov I see that this breaks the tests, likely because it will escape everything, including tags which should be there. I think I will defer to you on this one as I am not quite sure how all the logging components interact.

o-smirnov:
Thanks @JSKenyon. I think I fixed it correctly now -- the escape inside the emit() should just be a backstop for misbehaved code that's left markup in the message. And I've added an escape() at the appropriate place.

JSKenyon:
Ah perfect - I wasn't sure where that place was. Works on my end (i.e. pipeline is happy).

o-smirnov (approved):
OK I'm happy.

o-smirnov:
Hmm why can't I designate @JSKenyon as a reviewer, any idea? He's admin on the repo, and in the stimelators team.

o-smirnov:
Ah probably because @JSKenyon initiated the PR.

Athanaseus (approved):
LGTM!

PR #268, opened 2024-03-27
Drop Python3.7 support - heavily deprecated upstream.

JSKenyon:
Python3.7 is becoming very difficult to support as it has been dropped by upstream packages. I suggest removing it before the 2.0 release.


o-smirnov (approved):
Yes please

PR #269, opened 2024-03-28
Attempt at making substitutions work inside dicts and lists.

### LINKED ISSUE #271: logs from nested recipe steps not propagated to console
Picked up in #270.

JSKenyon:
This PR aims to improve substitutions by supporting them inside compound types. This probably needs extensive testing, but the following should now work:
```yaml
inputs:
  foo:  # foo is Dict[str, Tuple[File, float, str]] type.
    bar: [=expression_with_substitution, 1, example]  
```
or equivalently:
```yaml
inputs:
  foo:   # foo is Dict[str, Tuple[File, float, str]] type.
    bar: =LIST(substitution, 1, "example")
```
This should also support more complicated cases e.g.:
```yaml
inputs:
  foo:   # foo is List[Dict[str, Tuple[File, float, str]]] type.
    bar: 
      - baz: [=expression_with_substitution, 1, example]
```

### DIFF
--- a/scabha/evaluator.py
+++ b/scabha/evaluator.py
@@ -625,6 +625,31 @@ class Evaluator(object):
                         params_out[name] = new_value
                         if corresponding_ns:
                             corresponding_ns[name] = new_value
+                elif isinstance(value, (dict, DictConfig)):
+                    params_out[name] = self.evaluate_dict(
+                        value,
+                        corresponding_ns,
+                        defaults,
+                        sublocation,
+                        raise_substitution_errors,
+                        recursive,
+                        verbose
+                    )
+                elif isinstance(value, list):
+                    params_out[name] = type(value)(
+                        [
+                            self.evaluate_dict(
+                                {"dummy": v},
+                                corresponding_ns,
+                                defaults,
+                                sublocation,
+                                raise_substitution_errors,
+                                recursive,
+                                verbose
+                            )["dummy"] for v in value
+                        ]
+                    )
+
         return params_out
 
 
--- a/scabha/evaluator.py
+++ b/scabha/evaluator.py
@@ -635,18 +635,18 @@ class Evaluator(object):
                         recursive,
                         verbose
                     )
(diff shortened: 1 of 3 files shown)


scabha/evaluator.py
```diff
@@ -625,6 +625,31 @@ def evaluate_dict(self, params: Dict[str, Any],
...
+                    params_out[name] = self.evaluate_dict(
+                        value,
+                        corresponding_ns,
+                        defaults,
+                        sublocation,
+                        raise_substitution_errors,
+                        recursive,
+                        verbose
+                    )
+                elif isinstance(value, list):
```
o-smirnov: @JSKenyon clever, clever. Maybe use ``(list, ListConfig)`` just in case?

o-smirnov: Even better:

```python
params_out[name] = type(value)([x for x in
    self.evaluate_dict({f"[{i}]": v for i, v in enumerate(value)}, ...).values()
   ])
```

...because then any nested error messages will report the location of the error sensibly as "[n]".

JSKenyon: Ah, awesome. Definitely better than a profusion of `dummies`!

o-smirnov: Actually no need for the extra list comprehension, right? This should work too: 

```python
type(value)(self.evaluate_dict({f"[{i}]": v for i, v in enumerate(value)}, ...).values())
```

JSKenyon: Ok, I have made a final change. Let me know if it appeals.

o-smirnov: Oh, also the ``evaluate_dict()`` call needs ``sublocation=sublocation + [name]``, so that errors are reported from the correct nested namespace.

JSKenyon: Thanks for making those changes @o-smirnov - was only going to get to it today.

Athanaseus (approved):

PR #276, opened 2024-04-16
fixes #275

### LINKED ISSUE #275: clickify_parameters should be consistent with policies.repeat, and deal with Tuples correctly
Curently a ``List[x]`` I/O is dealt with by parsing a single comma-separated command-line argument. The ``policies.repeat`` field provides other options for passing lists (https://stimela.readthedocs.io/en/latest/reference/policies.html), and this ought to be honoured when constructing a CLI.

Likewise for tuples.

The ``multiple`` and ``nargs`` options to ``click.argument()`` and ``click.option()`` will help do this correctly.

o-smirnov:
@landmanbester could you please test that this branch does not break pfb-clean CLIs?

The change is that ``policies.repeat`` is now properly honoured for List and Tuple types, to accommodate the different styles of passing multiple values. As [per documentation](https://stimela.readthedocs.io/en/latest/reference/policies.html):

* ``repeat: list`` uses ``--option X Y``
* ``repeat: repeat`` uses ``--option X --option Y``
* ``repeat: '[]'`` uses ``--option [X,Y]``
* ``repeat: ','`` (or any other separator string) uses ``--option X,Y``

There is no default for the repeat policy, and ``clickify_parameters`` will raise an error if this is not explicitly specified for List and Tuple types. This may perhaps be an annoyance, and I'm open to using a default instead, but which one?

### DIFF
--- a/scabha/schema_utils.py
+++ b/scabha/schema_utils.py
@@ -125,15 +125,35 @@ def nested_schema_to_dataclass(nested: Dict[str, Dict], class_name: str, bases=(
 
 _atomic_types = dict(bool=bool, str=str, int=int, float=float)
 
-def _validate_list(text: str, element_type, schema):
+def _validate_list(text: str, element_type, schema, sep=",", brackets=True):
     if not text:
         return schema.default
-    if text == "[]":
-        return []
-    if text[0] == "[" and text[-1] == "]":
+    if brackets:
+        if text == "[]":
+            return []
+        if text[0] != "[" or text[-1] != "]":
+            raise click.BadParameter(f"can't convert to '{schema.dtype}', missing '[]' brackets")
         text = text[1:-1]
     try:
-        return [element_type(x) for x in text.split(",")]
+        return list(element_type(x) for x in text.split(sep))
+    except ValueError:
+        raise click.BadParameter(f"can't convert to '{schema.dtype}'")
+
+def _validate_tuple(text: str, element_types, schema, sep=",", brackets=True):
+    if not text:
+        return schema.default
+    print(text, element_types)
+    if brackets:
+        if text == "[]":
+            return []
+        if text[0] != "[" or text[-1] != "]":
+            raise click.BadParameter(f"can't convert to '{schema.dtype}', missing '[]' brackets")
+        text = text[1:-1]
+    elems = text.split(sep)
+    if len(elems) != len(element_types):
+        raise click.BadParameter(f"can't convert to '{schema.dtype}', tuple length mismatch")
+    try:
+        return tuple(element_type(x) for x, element_type in zip(elems, element_types))
     except ValueError:
         raise click.BadParameter(f"can't convert to '{schema.dtype}'")
 
@@ -148,23 +168,22 @@ def clickify_parameters(schemas: Union[str, Dict[str, Any]]):
     if type(schemas) is str:
         schemas = OmegaConf.merge(OmegaConf.structured(Schema),
                                 OmegaConf.load(schemas))
-    else:
-        schemas = OmegaConf.merge(OmegaConf.structured(Schema),
-                                  dict(inputs=schemas.inputs, outputs=schemas.outputs))
 
     decorator_chain = None
-    inputs = Cargo.flatten_schemas(OrderedDict(), schemas.inputs, "inputs")
-    outputs = Cargo.flatten_schemas(OrderedDict(), schemas.outputs, "outputs")
+    inputs = Cargo.flatten_schemas(OrderedDict(), getattr(schemas, 'inputs', {}), "inputs")
+    outputs = Cargo.flatten_schemas(OrderedDict(), getattr(schemas, 'outputs', {}), "outputs")
     for io in inputs, outputs:
         for name, schema in io.items():
             # skip outputs, unless they're named outputs
-            if io is schemas.outputs and not (schema.is_file_type and not schema.implicit):
+            if io is outputs and not (schema.is_file_type and not schema.implicit):
                 continue
 
             name = name.replace("_", "-").replace(".", "-")
             optname = f"--{name}"
             dtype = schema.dtype
             validator = None
+            multiple = False
+            nargs = 1
 
             # sort out option type. Atomic type?
             if dtype in _atomic_types:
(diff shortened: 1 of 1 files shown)


scabha/schema_utils.py
```diff
@@ -125,15 +125,35 @@ def nested_schema_to_dataclass(nested: Dict[str, Dict], class_name: str, bases=(
...
     try:
-        return [element_type(x) for x in text.split(",")]
+        return list(element_type(x) for x in text.split(sep))
+    except ValueError:
+        raise click.BadParameter(f"can't convert to '{schema.dtype}'")
+
+def _validate_tuple(text: str, element_types, schema, sep=",", brackets=True):
+    if not text:
+        return schema.default
+    print(text, element_types)
```
JSKenyon: Debugging relic.

scabha/schema_utils.py
```diff
@@ -125,15 +125,35 @@ def nested_schema_to_dataclass(nested: Dict[str, Dict], class_name: str, bases=(
...
+    if brackets:
+        if text == "[]":
+            return []
+        if text[0] != "[" or text[-1] != "]":
+            raise click.BadParameter(f"can't convert to '{schema.dtype}', missing '[]' brackets")
+        text = text[1:-1]
+    elems = text.split(sep)
+    if len(elems) != len(element_types):
```
JSKenyon, line 152: I have not checked, but aren't there two cases for tuples? `Tuple[Type1, Type2, ...]` and `Tuple[Type]` - does this check work for both?

JSKenyon (changes requested):
This looks fine to me bar the debugging relic but I will leave it to Landman to test in earnest.

landmanbester:
Sorry, got distracted by FFTs. Something is not working. If I have the following in a config file
```yaml
inputs:
  ms:
    dtype: List[URI]
    required: true
    abbreviation: ms
    info:
      Path to measurement set
 
outputs:
    {}
```
I now get
```
(pfbtest) ╭─landman@LAPTOP-ED4N50Q2 ~/software
╰─➤  pfb --help
Traceback (most recent call last):
  File "/home/landman/venvs/pfbtest/bin/pfb", line 33, in <module>
    sys.exit(load_entry_point('pfb-clean', 'console_scripts', 'pfb')())
  File "/home/landman/venvs/pfbtest/bin/pfb", line 25, in importlib_load_entry_point
    return next(matches).load()
  File "/usr/lib/python3.10/importlib/metadata/__init__.py", line 171, in load
    module = import_module(match.group('module'))
  File "/usr/lib/python3.10/importlib/__init__.py", line 126, in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
  File "<frozen importlib._bootstrap>", line 1050, in _gcd_import
  File "<frozen importlib._bootstrap>", line 1027, in _find_and_load
  File "<frozen importlib._bootstrap>", line 1006, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 688, in _load_unlocked
  File "<frozen importlib._bootstrap_external>", line 883, in exec_module
  File "<frozen importlib._bootstrap>", line 241, in _call_with_frames_removed
  File "/home/landman/software/pfb-clean/pfb/workers/main.py", line 10, in <module>
    from pfb.workers import (init, grid, degrid,
  File "/home/landman/software/pfb-clean/pfb/workers/init.py", line 21, in <module>
    @clickify_parameters(schema.init)
  File "/home/landman/venvs/pfbtest/lib/python3.10/site-packages/scabha/schema_utils.py", line 218, in clickify_parameters
    raise SchemaError(f"list-type parameter '{name}' does not have a repeat policy set")
scabha.exceptions.SchemaError: list-type parameter 'ms' does not have a repeat policy set
```
Should I change something in the config?

landmanbester:
Adding the following
```yaml
inputs:
  ms:
    dtype: List[URI]
    required: true
    abbreviation: ms
    info:
      Path to measurement set
    policies:
      repeat: '[]'
```
fixes the problem but it is a little annoying having to add it for every parameter. I kind of like `repeat: list` as the default, seems the most natural

scabha/schema_utils.py
o-smirnov: Ah thanks!

scabha/schema_utils.py
o-smirnov: Good question. [From my reading of this](https://docs.python.org/3/library/typing.html#annotating-tuples), ``Tuple[Type]`` means a single-element tuple, so this check should sill be valid.

As I just learned, Python recognizes ``Tuple[Type,...]`` as representing a variable-length tuple with a typed element. But for the purposes of specifying a CLI (both when passing arguments, and when parsing arguments), this seems to me to be indistinguishable from ``List[Type]``, so I'm content to not support the former.

o-smirnov:
> I kind of like repeat: list as the default, seems the most natural

Not really. That would be ``--option X1 X2 X3 ...`` on the command line -- click doesn't even support that for dashed options, because it becomes poorly defined w.r.t. whether X2 or X3 refers to the option, or to an extra positional argument. 

I would prefer ``repeat: repeat`` (``--option X1 --option X2``) or ``repeat: ','`` (``--option X1,X2``) as the default. Pick one?

Note that it's a a completely separate question from what should be the default when going in the opposite direction (converting a stimela params dict **to** CLI arguments). Currently, there is a hard check for an explicit repeat policy [here](https://github.com/caracal-pipeline/stimela/blob/aed211e11fdb8ec15be0a7814bea957df44dec25/stimela/kitchen/cab.py#L312). I would argue that this should stay in place.

Which raises a related question. Cabs have an optional cab-level ``policies`` section, which sets the default policies for all the parameters of the cab. I suspect this the reason the pfb-clean cabs have worked without an explicit repeat policy for each List-type parameter. Going in the opposite direction, via ``clickify_parameters()``, there is no way to specify default policies. Should we add the possibility of specifying such defaults?

scabha/schema_utils.py
JSKenyon: Ah yeah - thanks for understanding what I meant rather than what I typed. That makes sense to me!

JSKenyon (dismissed):

landmanbester:
> click doesn't even support that for dashed options, because it becomes poorly defined w.r.t. whether X2 or X3 refers to the option, or to an extra positional argument.

Guess I'm blind to this because of my aversion to positional arguments. It is very handy that clickify_parameters makes it possible to pass in arbitrary length lists. 

> I would prefer repeat: repeat (--option X1 --option X2) or repeat: ',' (--option X1,X2) as the default. Pick one?

I would prefer the latter but it's not a very strong preference.

> Should we add the possibility of specifying such defaults?

This would be handy. Maybe as a keyword argument to clickify_parameters? Alternatively, I guess I could also add it to [this file](https://github.com/ratt-ru/pfb-clean/blob/main/pfb/parser/schemas.py) if it's easier?

o-smirnov:
All right, I also vote ``repeat: ','`` (``--option X1,X2``) as the default. I shall make it thus.

> This would be handy. Maybe as a keyword argument to clickify_parameters? 

I shall make it thus.

landmanbester:
Awesome, thanks. Once you have it going I will test the changes in earnest and finish my review

o-smirnov:
All right, please test and re-review so we can merge this.

scabha/schema_utils.py
```diff
@@ -125,46 +125,79 @@ def nested_schema_to_dataclass(nested: Dict[str, Dict], class_name: str, bases=(
...
     for io in inputs, outputs:
         for name, schema in io.items():
             # skip outputs, unless they're named outputs
-            if io is schemas.outputs and not (schema.is_file_type and not schema.implicit):
+            if io is outputs and not (schema.is_file_type and not schema.implicit):
                 continue
 
+            policies = OmegaConf.merge(default_policies, schema.policies)
+            # impose default repeat policy of using a single argument for a list, i.e. X1,X2,X3
+            if policies.repeat is None:
```
landmanbester, line 192: Thanks for this one, no need to change anything in pfb even

scabha/schema_utils.py
```diff
@@ -175,16 +208,50 @@ def clickify_parameters(schemas: Union[str, Dict[str, Any]]):
...
+                        validator = lambda ctx, param, value, etype=dtype, schema=schema: \
+                            _validate_tuple(value, element_types=elem_types, schema=schema,
+                                           sep=policies.repeat, brackets=False)
+                    else:
+                        raise SchemaError(f"tuple-type parameter '{name}' does not have a repeat policy set")
+                else:
+                    # anything else will be just a string
+                    dtype = str
 
+            # if name == 'catalogs':
```
landmanbester: Did you mean to leave this in here?

landmanbester (dismissed):
All seems to be working in pfb, thanks. Just that one stray comment that you might not want to leave in there

o-smirnov:
Thanks @landmanbester. Please reapprove, my last commit invalidated your review.

landmanbester (approved):

PR #281, opened 2024-04-24
Some tiny fixes and improvements for kube backend + dask clusters.

JSKenyon:
This PR includes:

- Propagation of environment variables specified in `kube` backend options into the `dask_cluster` worker and scheduler pods. This is needed to support e.g. `NUMBA_CACHE_DIR` when using a `dask_cluster`.
- Fix for a log message which was causing crashes as it referenced a potentially undefined variable.
- Replaces the `dockerhub` busybox images with the `quay.io` equivalents to circumvent rate limits (which come into play when pulling images onto hundreds of nodes).

### DIFF
--- a/stimela/backends/kube/daskjob.py
+++ b/stimela/backends/kube/daskjob.py
@@ -95,6 +95,8 @@ def split_args(args):
 def daskjob_template(args):
     labels = dict(args.labels)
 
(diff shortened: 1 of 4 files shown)


o-smirnov (approved):
Great, thanks!

PR #285, opened 2024-04-30
fixes #284

### LINKED ISSUE #284: when a recipe-level alias specifies a default, this needs to override the cab default or step setting
Example in breifast. This:

```yml
    boost-threshold:
      aliases: [procres.boost-threshold]
      default: 2 
```

...doesn't take.

o-smirnov:
@landmanbester would like your input on this especially, as it tweaks the logic of aliases.

The major change is that aliases for which a value is set inside a step definition (or by the cab default) now propagate up to recipe-level defaults. And they are no longer shown as "Hidden" by ``stimela doc``.

### DIFF
--- a/stimela/kitchen/recipe.py
+++ b/stimela/kitchen/recipe.py
@@ -563,13 +563,27 @@ class Recipe(Cargo):
                 # define schema based on copy of the target, but preserve default
                 io[alias_name] = copy.copy(schema)
                 alias_schema = io[alias_name] 
-                # check that 
-                # default set from own schema, ignoring parameter setting 
+                # if default set in recipe schema, ignore any parameter setting in the step 
                 if orig_schema is not None and orig_schema.default is not UNSET:
-                    # also clear parameter setting to propagate our default
                     if step_param_name in step.params:
                         del step.params[step_param_name]
                     alias_schema.default = orig_schema.default
+                # else check if explicit value or a default is specified in the step -- make it the recipe default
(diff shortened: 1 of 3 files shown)


JSKenyon (approved):

stimela/kitchen/recipe.py
```diff
@@ -563,13 +563,27 @@ def _add_alias(self, alias_name: str, alias_target: Union[str, Tuple],
...
+                    elif step_param_name in step.cargo.defaults:
+                        defval = step.cargo.defaults[step_param_name]
+                    else:
+                        defval = schema.default
+                    if defval is not UNSET:
+                        alias_schema.required = False
+                        alias_schema.default = defval
+                        ## see https://github.com/caracal-pipeline/stimela/issues/284. No longer convinced
+                        ## these parameters should be marked as Hidden. After all, the recipe explicitly specifies them!
+                        ## mark it as hidden -- no need to expose parameters that are internally set this way
```
landmanbester, line 584: Are you leaving this here for posterity?

landmanbester (approved):
Seems sensible to me. Agree that aliases which have been set explicitly should not be marked as hidden

stimela/kitchen/recipe.py
o-smirnov: For a bit, yes. Want to revisit this after a bit so rather leave my thoughts in.

PR #298, opened 2024-05-19
make Py11 compliant

SpheMakh:
- Add `basetypes.EmptyClassDefault()` method because `dataclasses` in python 3.11 cannot have mutable (classes in this case) as default values. Updated all the class defaults I could fine.  It may be worth considering a 
```python
def __default__(self): 
    return dataclass.field(default_factory=self)
```
 method to our dataclasses then using ` foo: Bar = Bar.__default__()` when setting defaults.

- Save CASA code in a `tempfile` instead of parsing a multi-line code string; this did not work for me.

### DIFF
--- a/.github/workflows/python-package.yml
+++ b/.github/workflows/python-package.yml
@@ -16,7 +16,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python-version: ["3.8", "3.9", "3.10"]
+        python-version: ["3.8", "3.9", "3.10", "3.11"]
 
     steps:
     - uses: actions/checkout@v3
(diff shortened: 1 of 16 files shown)


o-smirnov (commented):
One last thing @SpheMakh -- have you tried this with CASA<6 (i.e. python2 based)? Just to make sure that code still runs.

o-smirnov:
Reminder @SpheMakh -- could you please check this implementation works with CASA5?

sphemakh:
> Reminder @SpheMakh -- could you please check this implementation works with CASA5?

It creates even more parsing problems. But this latest solution works for CASA >=4.7, 5, and 6

o-smirnov:
Ha, so simple in the end! All right, let's merge.

o-smirnov (approved):
LGTM.

PR #311, opened 2024-06-09
fixes #309. Fixes #310

### LINKED ISSUE #309: skip: =IFSET(recipe.foo) doesn't work as expected
Causes a skip if foo is NOT set (because =IFSET returns UNSET in this syntax and the skip logic doesn't check for this value).

### LINKED ISSUE #310: assign does not allow assigning dicts
The assignment dict is flattened (as for parameters), which is arguably not appropriate.

o-smirnov:


JSKenyon (approved):
This looks uncontroversial to me, provided that removing the flatten doesn't have unforeseen consequences.

o-smirnov:
It does have consequences for hierarchical assignments i.e.

```
assign:
  log: 
    dir: foo
```

might not work as expected anymore (has to be ``assign: log.dir: foo``) 

I don't think we have recipes in the wild using the former.

o-smirnov:
P.S. I do feel the ``assign`` logic overall is still a little unsettled in places, but I'm happy to muddle forward and let us refine it with more experience.

PR #318, opened 2024-06-13
Fix for bad shlex in venv activation

JSKenyon:
This is a minor fix which probably should reach master. Fixes a bug which results in the entire command being interpreted as a single path.


o-smirnov (approved):
LGTM, thanks!

PR #319, opened 2024-06-14
parse element type as function option

SpheMakh:


o-smirnov (approved):
Ah yes. The dreaded local scope binding bug. I always make that mistake!

PR #333, opened 2024-09-04
Add option to dump the recipe config file to the stimela run command  

landmanbester:
Seems to work as advertised. Not sure if this should be true by default. It may be useful for reporting issues. What do you think @o-smirnov ?


o-smirnov (approved):
LGTM!

o-smirnov:
Great, thanks! My initial take is that false by default is preferable, but we can always reconsider it down the line.

PR #343, opened 2024-11-07
Fix the regular expression escapes for more recent python versions. 

tmolteno:
The following syntaxWarnings appear in python 3.12.

```
scabha/basetypes.py:73: SyntaxWarning: invalid escape sequence '\w'
  match = re.fullmatch("((\w+)://)(.*)", value)

stimela/config.py:275: SyntaxWarning: invalid escape sequence '\$'
  env={key: value.replace('${', '\${') for key, value in os.environ.items()})

scabha/configuratt/resolvers.py:221: SyntaxWarning: invalid escape sequence '\['
  match = re.match("^(.*)\[(.*)\]$", incl)

stimela/backends/kube/__init__.py:198: SyntaxWarning: invalid escape sequence '\['
  event_format:  str = "=NOSUBST('\[k8s event type: {event.type}, reason: {event.reason}] {event.message}')"

scabha/cargo.py:527: SyntaxWarning: invalid escape sequence '\['
  attrs and info.append(f"[dim]\[{rich.markup.escape(', '.join(attrs))}][/dim]")

stimela/kitchen/recipe.py:509: SyntaxWarning: invalid escape sequence '\('
  if re.match('^\(.+\)$', step_spec):

stimela/kitchen/wranglers.py:207: SyntaxWarning: invalid escape sequence '\d'
  elif re.fullmatch('\d+', group):

stimela/commands/run.py:49: SyntaxWarning: invalid escape sequence '\w'
  match2 = re.fullmatch("^([\w.]+)::(.+)$", filename)

stimela/commands/run.py:87: SyntaxWarning: invalid escape sequence '\w'
  match2 = re.fullmatch("^([\w.]+)::(.+)$", filename)
```

The strings should be raw strings eg, `"((\w+)://)(.*)"` should become  `r'((\w+)://)(.*)'`, otherwise the \w will be treated as an escaped unicode character. Alternatively replace \w with \\w.

See [here](https://stackoverflow.com/questions/50504500/deprecationwarning-invalid-escape-sequence-what-to-use-instead-of-d) for details.

### DIFF
--- a/scabha/basetypes.py
+++ b/scabha/basetypes.py
@@ -70,7 +70,7 @@ class URI(str):
 
         If expand_user is True, ~ in (file-protocol) paths will be expanded.
         """
-        match = re.fullmatch("((\w+)://)(.*)", value)
+        match = re.fullmatch(r'((\w+)://)(.*)', value)
         if not match:
             protocol, path, remote = "file", value, False
         else:
--- a/scabha/cargo.py
+++ b/scabha/cargo.py
@@ -524,7 +524,7 @@ class Cargo(object):
                         attrs.append(f"choices: {', '.join(schema.choices)}")
                     info = []
                     schema.info and info.append(rich.markup.escape(schema.info))
-                    attrs and info.append(f"[dim]\[{rich.markup.escape(', '.join(attrs))}][/dim]")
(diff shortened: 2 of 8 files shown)


o-smirnov (approved):
LGTM, thanks for the fixes!

PR #351, opened 2024-12-02
Fix regexp for later versions of python

tmolteno:
Just a few more fixes, solved by converting regexp strings to raw strings.


SpheMakh (changes requested):
There are a few more instances that need to be fixed. These are the ones I found from the github search bar

1.  [re.match](https://github.com/search?q=repo%3Acaracal-pipeline%2Fstimela%20re.match&type=code)
2. [re.fullmatch](https://github.com/search?q=repo%3Acaracal-pipeline%2Fstimela+re.fullmatch&type=code)

tmolteno:
Have changed all regexp strings to raw strings, and modified the ones that had the necessary escape sequences in them.

SpheMakh (approved):
Looks good to me. Thanks Tim!

PR #359, opened 2024-12-17
Clickify cleanup

o-smirnov:
@landmanbester please check if this works with PFB.

I've removed the ``pass_missing_as_none`` logic from ``clickify_parameters()``, as upon closer investigation, this is what click now does always and anyways, except when dealing with some multiply-valued parameters (that is, Lists with a policy of repeat: repeat, in which case a missing parameter is passed as an empty tuple). I don't see any way to change this behaviour. The new code checks for UNSET defaults, and doesn't give click a default value for them, which results in a ``None`` being passed to the callable in the absence of the parameter. I think this behaviour is harmless and unlikely to trip us up, except when you have something perverted like

```python
def func(a: float=3):
  print(a)
```

with a schema that doesn't mention the default from the function signature:

```yaml
a:
  dtype: float
  required: false
```
...which results in ``func(a=None)`` when the command-line doesn't specify ``--a``. This is because neither stimela nor click knows anything about the signature actually, and has no choice but to trust what the schema says.

Note that ``pass_missing_as_none`` is still honoured by stimela itself when invoking a cab (i.e. is the difference between having nothing in ``**kw``, or having a ``None`` value in ``**kw``). Given your love for ``**kw``, I think this is what you expect, right?

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -15,7 +15,6 @@ packages = [
 
 [tool.poetry.dependencies]
 python = "^3.9"
-numpy = "^1.20"
 munch = "^2.5.0"
 omegaconf = "^2.1"
 importlib_metadata = { version = "4.13.0", python = "3.7" }
--- a/scabha/evaluator.py
+++ b/scabha/evaluator.py
@@ -3,7 +3,6 @@ import glob
 import os.path
 import fnmatch
 import pyparsing
-import numpy as np
 pyparsing.ParserElement.enable_packrat()
 from pyparsing import *
 from pyparsing import common
@@ -182,7 +181,7 @@ class FunctionHandler(ResultsHandler):
     
     def IS_NUM(self, evaluator, args):
         def is_num(x):
-            return np.isscalar(x) and (np.isreal(x) or np.isscalar(x)) 
+            return isinstance(x, (bool, int, float, complex)) 
         return self.evaluate_generic_callable(evaluator, "IS_NUM", is_num, args, min_args=1, max_args=1)
 
     def IF(self, evaluator, args):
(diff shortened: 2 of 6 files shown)


o-smirnov:
Run ``tests/scabha_tests/test_clickify.py`` to test this behaviour (and feel free to extend).

landmanbester (commented):
This seems ok to me. It will require handling optional lists differently in pfb but that should be doable. More worrying is the numpy dependency here https://github.com/caracal-pipeline/stimela/blob/36d1abb78273acd91fad3bedb85fbf09cf2a52b0/pyproject.toml#L18
Not sure if this relates to the current setuptools issues but even with setuptools[core] I get the following conflicts
```
ERROR: pip's dependency resolver does not currently take into account all the packages that are installed. This behaviour is the source of the following dependency conflicts.
codex-africanus 0.4.1 requires numpy<3.0,>=2.0, but you have numpy 1.26.4 which is incompatible.
dask-ms 0.2.23 requires numpy>=2.0.0, but you have numpy 1.26.4 which is incompatible.
```
What does stimela use numpy for anyway?

o-smirnov:
Oh fsck numpy, I was only using it in one place, and I can work around without it. I've taken the dependency out, try again.

> It will require handling optional lists differently in pfb but that should be doable. 

Well nothing's changed there -- it seems that click was passing optional lists as ``()`` all along, even with ``pass_missing_as_none`` in effect. So if you depend on ``optional_list is None`` and don't cope with empty lists, your CLI has always been slightly broken. But checking for an empty optional list or None with ``if optional_list:`` is probably harmless to do in most cases, right? Unless "empty list" and "None" are two distinct logic cases for you, which would be a bit perverse.

landmanbester (approved):
Seems to be working for me. Empty lists are still parsed as `None` for some reason though so this does not really change anything in pfb. I guess it's because I don't have a repeat policy set. Note I will be mostly AFK until I get back from leave in early Jan

o-smirnov:
OK well if @Athanaseus approves, I'll merge this and do a point release, so he can get his breizorro PR in as well.

PR #368, opened 2025-03-01
Enables list item syntax e.g. =current.foo[2]. Adds option to skip freshness checks

o-smirnov:


scabha/evaluator.py
JSKenyon, line 1: I assume that this only supports indexing a single item in a list? Is there a use-case for slicing i.e. `list[start:stop:step]`?

JSKenyon (approved):
Just one question, otherwise this looks good to go!

scabha/evaluator.py
o-smirnov: There may be, but that would mean adding more grammar to the parser (or else a built-in SLICE function), so maybe let's leave it until demand arises...

PR #370, opened 2025-03-02
fixes #304

### LINKED ISSUE #304: Consider changing the name of the "info" namespace to "self" or "this"
Similar to python, it reinforces the idea that the user is interacting with the current step

Other suggestions:

info.label -> self.name
info.suffix -> self.name_suffix

o-smirnov:
@JSKenyon would you please test with your basic selfcal workflow, changing ``info`` to ``self``? I've tested with TRON and it seems to work, but there the use of ``info`` is minimal.


JSKenyon (approved):
Looks uncontroversial.

o-smirnov:
As long as it works. My only mild concern was having two namespaces somehow become proper duplicates. The way I assign them to SubstitutionNS, ``info`` and ``self`` do refer to the same object, and we don't do deep copies anywhere, so this shouldn't be an issue.

PR #378, opened 2025-03-25
Fixes #377. Fixes #376.

### LINKED ISSUE #376: Fix backend settings docs
There are a few mistakes in https://stimela.readthedocs.io/en/latest/reference/backend_ref.html.

``backends`` -> ``backend``. Also check for instances of ``enable: true``, as this has been deprecated.

### LINKED ISSUE #377: add option to bind container directories to different paths
This need arose while dealing with https://github.com/caracal-pipeline/cult-cargo/issues/87. It's an example of a more generic need for cabs that depend on external data which may be either auto-downloaded, or perhaps provided in a static version by the system administrator.

Use case: I want to configure the cult-cargo python-astro image to keep the astropy cache in ``/var/cache/astropy``. The cargo images can ensure an up-to-date cache when built. However, I want to provide the user with the option to explicitly refresh the cache (and then use this cache in any other cabs that use astropy). This requires binding a local cache directory (e.g. ``~/.astropy``) to ``/var/cache/astropy`` inside the container.

The current ``backend.singularity.bind_dirs`` syntax implies identical source and target paths, i.e.

```yml
bind_dirs:
  /path: ro
  /other_path: rw
```

Suggest extending this to recognizing

```yml
bind_dirs:
  /path: ro
  /var/cache/astropy: ~/.astropy::rw
```

with the logic being, if the value contains ``::``, treat is as ``path::mode``, otherwise treat it as mode only, and assume the same pathname.

Even more useful would be to recognize {}-substitutions both inside the key and the value (I think only values are currently subject to substitutions), because it will be easier to write this "refresh astropy data" cab more robustly if I could use something like:

```yml
bind_dirs:
  /path: ro
  '{config.vars.cult-cargo.astropy-cache-dir}': '{current.cache-dir}::rw'
```

Easy to implement but any thoughts from @JSKenyon @SpheMakh @sjperkins @landmanbester first?

o-smirnov:
Other fixes:

Throw substitution errors when they arise in backend settings (this was meant to be done, but wasn't actually happening due to a bug).

Print/log more friendly command lines when running python and casa flavour cabs.

### DIFF
--- a/docs/source/reference/backend_ref.rst
+++ b/docs/source/reference/backend_ref.rst
@@ -37,7 +37,7 @@ Native backend settings
 
 The native backend has only a couple of settings::
 
-    backends:
+    backend:
         native:
             enable: true
             virtual_env: ~/venvs/my_venv
@@ -51,7 +51,7 @@ Singularity backend settings
 
 The Singularity backend has the following settings::
 
-    backends:
+    backend:
         singularity:
             enable: true
             image_dir: ~/.singularity
@@ -74,7 +74,7 @@ Slurm wrapper settings
 
 Slurm is a wrapper, not a backend per se. It can be used in combination with the native and Singularity backends to schedule steps as Slurm jobs (using ``srun``). Enabling it can be as simple as setting ``enable`` to true::
 
-    backends:
+    backend:
         slurm:
             enable: false
             srun_path:              # optional path to srun executable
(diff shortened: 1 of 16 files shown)


o-smirnov:
OK, tests failing, please hold off reviewing until I've had a look.

o-smirnov:
OK this is ready for review.

o-smirnov:
Hang on one more fix incoming.

o-smirnov:
@JSKenyon or @SpheMakh please review now.

JSKenyon (approved):
I haven't done any testing but it looks ok to me.

Issue #375, opened 2025-03-07
Add =SELECT() function


o-smirnov:
Have ugliness like this on the breifast power spectrum branch:

```yml
    procres:
      cab: breifast.process-residual-cube
      info: processes cube and generates candidate catalog
      params:
        cds: =(recipe.is_raw and recipe.cds) or 
              (recipe.is_convolved and steps.convolve.out-cds) or
              (recipe.is_fd and steps.forward-diff.out-cds) or 
              (recipe.is_ps and steps.power-spectrum.out-cds) or "unsupported timesscale setting {recipe.timescale}"
```

Underlying reason there's a multi-way selection of the input file based on which mode the step is invoked in. This could be simplified by having a built in select function. Something like

```
=SELECT(cond1, result1, 
        cond2, result2, 
         ...)
```

Logic being, evaluate each conditional in turn, and return the first result where the conditional is true, or UNSET if none are true.

o-smirnov:
Or ``=CASES()`` as @JSKenyon suggests.

o-smirnov:
Everyone good with ``=CASES()`` as described? I feel motivated to implement this quickly for 2.0.3, and to push out 2.0.3 (because it is needed for cult-cargo 0.2.0).

o-smirnov:
Should probably add an ``ERROR()`` function as well, as that last line should be ``ERROR("unsupported timescale setting {recipe.timescale}")``.

JSKenyon:
I am happy with cases - I think it is pretty clear. Having the ability to have an error case is super useful in general.

PR #385, opened 2025-03-28
Issue 384

o-smirnov:
This supercedes #381, but you can review them in turn as well @JSKenyon.

This adds a ``TRY()`` as discussed in #375.

I've also cleaned up handling of ``Unresolved`` values. Old pattern was ``type(x) is UNSET`` which only handled a subset of the unresolved-value conditions (``Unresolved`` is a parent class of ``UNSET`` for all unresolved conditions, see ``scabha/basetypes.py``).

### DIFF
--- a/scabha/basetypes.py
+++ b/scabha/basetypes.py
@@ -49,7 +49,7 @@ class UNSET(Unresolved):
     pass
 
 class Placeholder(Unresolved):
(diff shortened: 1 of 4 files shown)


JSKenyon (approved):
LGTM!

PR #390, opened 2025-04-02
Prep 2.1

o-smirnov:
As suggested on chat, the last release should really have been designated as 2.1 due to significant new features.

So, this PR is in prep for 2.1 proper (@JSKenyon has put in a quick bugfix PR in the meantime). 

I've basically taken the opportunity to update the docs (and include a link to the paper in them). Have a gander at https://stimela.readthedocs.io/en/prep-2.1

PR #394, opened 2025-04-05
Issue 391

o-smirnov:


o-smirnov:
This is based off #390 so maybe review that first.

o-smirnov:
@JSKenyon @SpheMakh please review.

scabha/cargo.py
```diff
@@ -126,9 +131,9 @@ class Parameter(object):
     aliases: Optional[List[str]] = ()
 
     # if true, create parent directories of file-type outputs if needed
-    mkdir: bool = True
-    # if true, create directory-type outputs themselves
-    mkdir_self: bool = False
+    mkdir_parent: bool = True
```
JSKenyon: My intuition is failing me on this change - can this break existing recipes? I almost wonder if this would be better as:
```python
mkdir: bool = True
mkdir_policy: str = "parents"   # With alternatives being "all", "child" etc. May also allow for more specific behaviour e.g. controlling exists_ok. 
```

stimela/kitchen/cab.py
```diff
@@ -228,17 +228,26 @@ def update_environment(self, subst):
...
+            # get skip setting
+            skip = self.get_schema_policy(schema, 'skip')
+            if skip is None and schema.implicit:
+                skip = self.get_schema_policy(schema, 'skip_implicits')
+            # skip if explicitly True
+            if skip:
+                continue
+            # skip non-named outputs, unless skip is explicitly False
+            if not (schema.is_input or schema.is_named_output) and skip is not False:
+                continue
```
JSKenyon, line 244: I don't understand this check. Is there reason to check `is not False` over `is True`?

o-smirnov: Yes, it can be ``None`` in which case the default logic (skip if not an input or a named output) must apply.

JSKenyon: What determines whether something is a named output - I think I have lost the thread there. Does that mean it is explicitly part of the output definition as opposed to an implicit output?

scabha/cargo.py
o-smirnov: The previous change (already in master and R2.0.3) already breaks existing recipes in the case where they have an output directory parameter, and this does not exist up front. These will need an ``mkdir: True`` setting now. Reminder of timeline:

* pre 2.0.3, we used to have a single ``mkdir`` property which was True by default, and applied both to the parent directory and the output itself. This caused pain for zarr and MS outputs (and other cases where the package expected to create its own directory). We had to specify ``mkdir: False`` all over the place to work around it. This in turn caused pain if the parent of the zarr did not exist.

* as of https://github.com/caracal-pipeline/stimela/commit/db8307423900e6f3c3cf22796d56b8bf156243f9 and 2.0.3, we introduced ``mkdir: True`` (meaning parents) and ``mkdir_self: False`` (meaning output itself) to decouple these two settings.

* I have since convinced myself that ``mkdir_self`` is a bit ugly and unintuitive, hence the current suggested change. It's only been three weeks so I'm sure nobody has adopted ``mkdir_self`` yet.

I could be convinced by ``mkdir_policy``... a bit of an extra complication but perhaps more future-proof.

stimela/kitchen/cab.py
o-smirnov: ``-o OUTFILE`` is a named output. It's something that's an output, but its filename is really a user-controlled input. As opposed to implicit outputs, which are named automatically.

scabha/cargo.py
o-smirnov: Since ``mkdir_parent: True`` is new and not yet adopted, we may as well recast it as ``mkdir_policy: parents``. The default setting is probably appropriate 99% of the time.

JSKenyon: The policy suggestion is just that - whatever you feel is more appropriate. Are custom dataclasses supported at this level i.e. could we have a `MkdirPolicy` object which exposed some options? I agree that the default is going to cover most usage, but as you say, it may be nice to future-proof it we can. There will always be another weird use-case.

stimela/kitchen/cab.py
JSKenyon: Got it.

scabha/cargo.py
o-smirnov: > The policy suggestion is just that - whatever you feel is more appropriate. Are custom dataclasses supported at this level i.e. could we have a `MkdirPolicy` object which exposed some options? 

Yes, exactly what I was thinking...

o-smirnov: Would propose a ``path_policies`` property which could also absorb ``access_parent_dir`` and ``write_parent_dir`` (only used in a couple of places in cult-cargo currently.)

JSKenyon: Sounds sensible to me if you are happy to do the implementing (I would offer but it would likely take me exponentially longer).

o-smirnov: OK @JSKenyon I have pushed proposed changes as discussed. This also adds deprecation warnings.

In the future, we can consider deprecating ``mkdir`` itself, and perhaps ``must_exist``, in favour of fields inside ``path_policies`` too. I just feel it's too much of a big change for now, as these two are used extensively (as opposed to the more obscure properties I have marked as deprecated). In any case the deprecation warning pattern provides a neat roadmap to doing so.

JSKenyon (approved):
Looks good to me!

PR #414, opened 2025-04-24
Add a LEN function to allow zip-like functionality

tmolteno:
Add a LEN function to scabha to allow loops with ZIP-like functionality.

I want to loop over two lists (similar to x,y = zip(x-list, y-list)). This can be achieved by using a LEN() function to return the length of an array. Ideally we'd add 

        for_loop:
                var: x,y
                over: x-list, y-list

But this seemed difficult. Instead this PR adds a LEN function to scabha, which enables zip-like functionality by passing an array of indexes around. Below is an example (also included as a test case in this PR)
```
cabs:
    echo:
        command: echo
        inputs:
            args:
                dtype: List[str]
                required: true
                policies:
                    positional: true
                    repeat: false
    help:
        command: sleep

    sleep:
        command: sleep
        inputs:
            seconds:
                dtype: int
                required: true
                policies:
                    positional: true


opts:
    log:
        dir: test-logs/logs-{config.run.datetime}
        nest: 3
        symlink: logs


loop_recipe:
    name: "loop recipe"
    info: 'test recipe with loops and sleeps'

    inputs:
        index-list: List[int]
        list-1: List[str]
        list-2: List[str]


    for_loop:
        var: i
        over: index-list
        display_status: "{var}={value} {index1}/{total}"

    steps:
        echo:
            cab: echo
            params:
                #args: [=recipe.x,=recipe.y]
                args: =LIST(recipe.list-1[recipe.i],recipe.list-2[recipe.i])

        slp1:
            cab: sleep
            params:
                seconds: 2

main_recipe:
    name: "main recipe"

    assign:
        list-1: ['test', 'me']
        list-2: ['x', 'y']

    steps:
        loop:
            recipe: loop_recipe
            params:
                index-list: =RANGE(LEN(recipe.list-1))
                list-1: =recipe.list-1
                list-2: =recipe.list-2
```


scabha/evaluator.py
```diff
@@ -176,6 +176,11 @@ def make_list(*x):
             return list(x)
         return self.evaluate_generic_callable(evaluator, "LIST", make_list, args)
 
+    def LEN(self, evaluator, args):
+        def make_len(*x):
```
o-smirnov: @tmolteno are you sure this works correctly? ``args`` is a list of function arguments of which the list is the first argument. Firstly, you need a check for just a single argument. Then, the function needs to be ``make_len(x)``. Otherwise it always returns 1 surely?

o-smirnov (changes requested):
Not sure args is correct...

scabha/evaluator.py
tmolteno: Oops. Caught that one, and added a test with zip-like functionality.

scabha/evaluator.py
```diff
@@ -176,6 +176,11 @@ def make_list(*x):
             return list(x)
         return self.evaluate_generic_callable(evaluator, "LIST", make_list, args)
 
+    def LEN(self, evaluator, args):
+        def make_len(x):
+            return len(x)
+        return self.evaluate_generic_callable(evaluator, "LEN", make_len, args)
```
o-smirnov: @tmolteno please add min_args=1, max_args=1 to this call, so that it throws an error message if things are otherwise.

o-smirnov (changes requested):
Just a tiny change to ensure errors are thrown, otherwise LGTM. Thanks!

scabha/evaluator.py
o-smirnov: @tmolteno reminder of requested change here.

o-smirnov (changes requested):

tmolteno:
Requested changes have been made. Tests (test_len_loop.yml) is passing.

o-smirnov (approved):
LGTM, thanks!

PR #417, opened 2025-05-07
fixed #416. Fixes redundant/incorrect labelling of errors in log

### LINKED ISSUE #416: allow function calls within GLOB() (and EXISTS and ERROR)
Mea culpa, I'm regretting the decision to allow GLOB(), EXISTS() and ERROR() to take "any sequence" as an argument. This broke when the argument contained an expression with parentheses.

Probably an easy fix to the parser though to try to recognize a proper expression first as a preference.

o-smirnov:


scabha/evaluator.py
```diff
@@ -278,7 +278,7 @@ def GLOB(self, evaluator, args):
         if isinstance(pattern, Unresolved):
             return pattern
         return sorted(glob.glob(pattern))
-
+
```
Athanaseus: Extra whitespace introduced here
```suggestion

```

Athanaseus (dismissed):
LGTM!

scabha/evaluator.py
o-smirnov: Thanks, pushed change, could you please re-review @Athanaseus?

Athanaseus (approved):

PR #419, opened 2025-05-19
Add more fully featured custom console print.

JSKenyon:
Currently the `### running command...` log messages are difficult to copy and paste due to the inclusion of newline characters. This PR removes wrapping/newline characters from the relevant messages as well as makes them more obvious in the logs. As part of this fix, messages may include a `custom_console_print=True` which means the message will not be handled by the `rich.logging.RichHandler.emit` method and instead be printed directly using the `self._console.print` method. The new code will forward all known arguments to the `self._console.print` method from the extras included on the record object.

### DIFF
--- a/stimela/stimelogging.py
+++ b/stimela/stimelogging.py
@@ -21,6 +21,24 @@ from warnings import warn
 
 from . import task_stats
 
+CONSOLE_PRINT_OPTIONS = [
+    "sep",
+    "end",
+    "style",
+    "justify",
+    "overflow",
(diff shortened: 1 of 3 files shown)


JSKenyon:
An example of the updated output:
image

JSKenyon:
I have updated the tests to be aware of the new formatting. Let me know if you would prefer something else @o-smirnov.

o-smirnov (approved):
LGTM, thanks!

PR #420, opened 2025-05-19
Add option to singularity backend to retain temporary directories.

JSKenyon:
This adds the option to retain the temporary directories created by stimela when launching singularity containers. This is useful when debugging as otherwise the CLI invocations become stale. Note that the custom class is temporary - once stimela is strictly >= Python3.12, we can go back to using `tempfile.TemporaryDirectory(delete=False)`.

This PR does not fix issues relating to temporary directories on slurm clusters.


o-smirnov (approved):
LGTM!

PR #421, opened 2025-05-20
Attempt to update and improve URI handling.

JSKenyon:
This PR is my attempt at improving URI handling. I was running into problems when specifying an `MS` type input via the command line when including relative paths e.g. `~/path/to/ms`. The existing URI class is a little odd in that it subclasses `str`, but doesn't update its `__repr__` or `__str__` methods so although the URI is parsed correctly, calling `str(uri_obj)` will always return the input string. This results in parameters entering the command with unresolved relative paths which is problematic as `$HOME` may not be the same inside and outside the container. This PR attempts to overhaul the base URI class to make use of `uritools`. This should be more robust and leave room for more advanced functionality in the future. Note that I have removed the `parse` method in favour of a more comprehensive `__init__` (which I believe is harmless as I cannot see any invocations of `URI.parse` in the code). Finally, I have removed the optional `expand_user` option - this appeared to be unused and I cannot think of a situation in which we would want to disable this functionality (but feel free to disagree).

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -25,6 +25,7 @@ psutil = "^5.9.3"
 rich = "^13.7.0"
 dill = "^0.3.6"
 typeguard = "^4.2.1"
+uritools = "^5.0.0"
 
 [tool.poetry.scripts]
 stimela = "stimela.main:cli"
--- a/scabha/basetypes.py
+++ b/scabha/basetypes.py
@@ -68,15 +68,30 @@ class URI(str):
         self.query = uri_components.query
         self.fragment = uri_components.fragment
 
+        # NOTE(JSKenyon): We assume that remote URIs are properly formed and
+        # absolute i.e. we do not reason about relative paths for e.g. s3. The
+        # following attempts to express paths relative to the cwd but will
+        # prefer absolute paths when inputs are outside the cwd. This can be
+        # changed when stimela's minimum Python >= 3.12 by using the newly
+        # added `walk_up` option.
         if self.scheme == "file":
-            self.path = str(Path(uri_components.path).expanduser().absolute())
+            cwd = Path.cwd().absolute()
+            abs_path = Path(uri_components.path).expanduser().resolve()
+            self.abs_path = str(abs_path)
+            try:
+                self.path = str(abs_path.relative_to(cwd))
+            except ValueError as e:
+                if "is not in the subpath" in str(e):
+                    self.path = self.abs_path
+                else:
+                    raise e
         else:
-            self.path = uri_components.path
+            self.path = self.abs_path = uri_components.path
 
         self.full_uri = uritools.uricompose(
             scheme=self.scheme,
             authority=self.authority,
-            path=self.path,
+            path=self.abs_path,
             query=self.query,
             fragment=self.fragment
         )
--- a/scabha/basetypes.py
+++ b/scabha/basetypes.py
@@ -10,6 +10,8 @@ from typeguard import (
     check_type, TypeCheckError, TypeCheckerCallable, TypeCheckMemo, checker_lookup_functions
 )
 from inspect import isclass
+import uritools
+from pathlib import Path
 
 
 def EmptyDictDefault():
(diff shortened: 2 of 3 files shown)


JSKenyon:
This currently breaks the tests as paths are fully resolved. I will wait for input from @o-smirnov before making further changes as I may have missed some subtlety that makes leaving paths unresolved preferable.

o-smirnov:
This is all to the good. I don't think I put a lot of thought into the ``URI`` class at the time, so I'm glad you're cleaning it up.

The failing tests (that I can see) are just due to paths being printed as fully resolved, and not matching the expected output, correct? By itself this seems trivial, but I understand your caution. I think I tried to keep them unresolved because (a) the aesthetics of less cluttered command lines, and (b) having a vague hunch that it would be better to let the OS resolve things, at least relative to the current directory,

I could have completely thought myself into a knot over (b) though so let's discuss. Thinking out loud, if $HOME is different inside the container / over slurm, should we not be delaying the resolving?..

Regarding (a), maybe we should resolve, then check if the resulting path is a child of CWD, and strip the CWD back off if so. This will declutter the path in 95% of the cases, and should be fairly robust, no?

Finally, speaking of aesthetics, there's something that bothers me about the "---INVOKING---" line, but I can't quite put my finger on it yet. Let me think, maybe we tweak it as part of this PR. Maybe something more succinct, like "$" before the command (using a console highlight)? I can see the convenience of having the command on a line all by itself though, easily selectable for cut-and-paste...

JSKenyon:
> The failing tests (that I can see) are just due to paths being printed as fully resolved, and not matching the expected output, correct? By itself this seems trivial, but I understand your caution.

Yeah, in principle these should be easy to fix. I wanted to confirm whether the new behaviour is desirable first.

> I think I tried to keep them unresolved because (a) the aesthetics of less cluttered command lines, and (b) having a vague hunch that it would be better to let the OS resolve things, at least relative to the current directory,

I can definitely see the value of (a), but (b) is tougher. 

> I could have completely thought myself into a knot over (b) though so let's discuss. Thinking out loud, if $HOME is different inside the container / over slurm, should we not be delaying the resolving?..

I will run a check today to confirm my suspicions about `$HOME`. I think the important point is that a user shouldn't be reasoning about paths inside the containers i.e. there should be no friction in switching between the native and singularity backends. Using absolute paths has the advantage of eliminating uncertainty, and should be completely safe provided that the inputs are mounted correctly. I am a little nervous about paths resolving differently inside and outside containers.
 
> Regarding (a), maybe we should resolve, then check if the resulting path is a child of CWD, and strip the CWD back off if so. This will declutter the path in 95% of the cases, and should be fairly robust, no?

Provided CWD is always the same inside and outside the container. In principle, this should be relatively easy with `pathlib`. Might take a stab at it.

> Finally, speaking of aesthetics, there's something that bothers me about the "---INVOKING---" line, but I can't quite put my finger on it yet. Let me think, maybe we tweak it as part of this PR. Maybe something more succinct, like "$" before the command (using a console highlight)? I can see the convenience of having the command on a line all by itself though, easily selectable for cut-and-paste...

Yeah, I wasn't 100% sold on the formatting so I am open to ideas. I just like leaving obvious signposts for the user. Maybe the fact that the header text is centered looks awkward?

JSKenyon:
Hmm, the plot thickens. This is the command which causes problems with stimela:
```
/usr/local/bin/singularity exec --pwd /home/kenyon/git_packages/lightcurve-pipeline/lightcurve_pipeline/flagging_strategies --contain --bind /home/kenyon/git_packages/lightcurve-pipeline/lightcurve_pipeline/flagging_strategies:/home/kenyon/git_packages/lightcurve-pipeline/lightcurve_pipeline/flagging_strategies:rw --bind /home/kenyon/reductions/prox_cen/1616737569_sdp_l0-Proxima_Cen-corr.ms:/home/kenyon/reductions/prox_cen/1616737569_sdp_l0-Proxima_Cen-corr.ms:rw --bind /tmp/tmpd0biypvd:/tmp:rw /home/kenyon/.singularity/quay.io-stimela2-tricolour:cc0.2.0.simg tricolour --config /home/kenyon/git_packages/lightcurve-pipeline/lightcurve_pipeline/flagging_strategies/target_flagging_1_narrow.yaml --flagging-strategy polarisation --data-column DATA --disable-post-mortem ~/reductions/prox_cen/1616737569_sdp_l0-Proxima_Cen-corr.ms
```

Notice the relative path at the end. The weird thing is that if I run that command directly i.e. outside of stimela, there is no error. The error occurs without the changes I have made. Now I am really confused.

JSKenyon:
Ah, the above will fail if the path includes string quotes e.g.:
```
/usr/local/bin/singularity exec --pwd /home/kenyon/git_packages/lightcurve-pipeline/lightcurve_pipeline/flagging_strategies --contain --bind /home/kenyon/git_packages/lightcurve-pipeline/lightcurve_pipeline/flagging_strategies:/home/kenyon/git_packages/lightcurve-pipeline/lightcurve_pipeline/flagging_strategies:rw --bind /home/kenyon/reductions/prox_cen/1616737569_sdp_l0-Proxima_Cen-corr.ms:/home/kenyon/reductions/prox_cen/1616737569_sdp_l0-Proxima_Cen-corr.ms:rw --bind /tmp/tmp8jfh0md7:/tmp:rw /home/kenyon/.singularity/quay.io-stimela2-tricolour:cc0.2.0.simg tricolour --config /home/kenyon/git_packages/lightcurve-pipeline/lightcurve_pipeline/flagging_strategies/target_flagging_1_narrow.yaml --flagging-strategy polarisation --data-column DATA --disable-post-mortem "~/reductions/prox_cen/1616737569_sdp_l0-Proxima_Cen-corr.ms"
```
I suspect this is the root issue although I don't yet know the cause.

o-smirnov:
Not surprised, quotes block tilde expansion. But how did quotes get into the path?

JSKenyon:
> Not surprised, quotes block tilde expansion. But how did quotes get into the path?

Digging at present. The quotes definitely aren't in the command which we print in the log but the error message is consistent when I include them manually. Trying to see exactly what goes into the asyncio calls.

JSKenyon:
Ok, I have pushed some changes. In principle, relative paths will be retained when they are inside the current working directory whereas absolute paths will be preferred for things outside. This is a limitation of `relative_to` in Python <= 3.12. I had to make one small change to the tests as technically a relative URI is a bit unconventional. Relative file URIs are should have the following form `file:path/relative/to/cwd` whereas absolute file URIs are either `file:/an/absolute/path` or `file:///an/absolute/path`.

o-smirnov (approved):
LGTM!

PR #423, opened 2025-05-28
Issue 422

### LINKED ISSUE #422: Add a construct to support defining paths relative to a recipe file
Currently, there is no way to tell stimela to look for an input relative to the location of a recipe. As an example, assume that the current working directory is `foo` and my recipe is in `foo/bar/recipe.yml`. The recipe requires a data file e.g. `strategy.yml` which also lives in `foo/bar`. If, from `foo`, I call `stimela run bar/recipe.yml`, it will not find the data file as its path will be interpreted relative to the current working directory rather than the recipe which requires it.

The solution is probably the addition of something along the lines of a `recipe.recipe_dir` construct such that the data file input could be specified as `{recipe.recipe_dir}/strategy.yml`.

o-smirnov:


o-smirnov:
P.S. Deleted code in [scabha/configuratt/cache.py](https://github.com/caracal-pipeline/stimela/pull/423/files#diff-e7a0e09e5511a9e76c464cd638dad7c430e70dd8771b6d460cafc3ccd0e9f982) is just cleanup of dead code -- these functions were migrated into a different module at one point, and I'd neglected to delete the old copies.

JSKenyon (approved):
Looks good and has been working for me.

PR #426, opened 2025-06-02
escape omegaconf interpolation in git info string

o-smirnov:
This is a band-aid for a more pernicious problem (though perhaps not so pernicious, seeing as it's been there forever and nobody tripped over it)...

I had referenced "${}-interpolation" in my git commit message. The dependency dict that stimela accumulates inserts the git commit message into the dict -- the "${}" then tripped an error in the interpolation lexer. :facepalm:

This fix applies an escape string -- but more generally, text containing "${}" will trip strange errors when assigned to an OmegaConf dict. It might be a helluva job to sanitize every assignment...

### DIFF
--- a/scabha/configuratt/deps.py
+++ b/scabha/configuratt/deps.py
@@ -124,7 +124,7 @@ class ConfigDependencies(object):
         for line in branches.decode().split("\n"):
             line = line.strip()
             if line.startswith("*"):
-                gitinfo.branch = line[1:].strip()
+                gitinfo.branch = line[1:].strip().replace("${", "\\${")
                 break
         # get description
         try:


JSKenyon (approved):
Sensible bandaid. I guess this all stems from the fact that OmegaConf attempts to interpolate `${}` constructs. I see that is is possible to disable the built-in resolvers (https://omegaconf.readthedocs.io/en/latest/custom_resolvers.html#clear-resolver). I wonder if it is worth clearing all resolvers bar the newly added one?

JSKenyon:
Although I don't know what the fallback behaviour is in OmegaConf.

o-smirnov:
Yeah I could disable the resolvers (and I see I did at one point -- there's some code commented out doing that...), but all that does is skip more exotic things like ``${oc:foo}``. I don't see a way to disable ``${}`` interpolation globally. Nor would we want to, given that we've just added a feature that relies on it.

I don't think it's a trainsmash. Just that the error messages it yields are truly incomprehensible. Especially if the interpolation is triggered by an assignment to a dict within the Stimela code itself (as opposed to when loading external YAML).

PR #432, opened 2025-06-17
fixes #431

### LINKED ISSUE #431: nested assign_based_on fails in prevalidation of a step when value is a formula
In prevalidation formulas are unresolved, so assign_based_on where the base value is a formula fails. Reported by @JSKenyon.

o-smirnov:


JSKenyon (approved):
Works for me on both the reproducer and the real example I was working on. What does the `abo` in `ignore_abo_errors` stand for?

o-smirnov:
"assign based on" :)

PR #437, opened 2025-06-19
Make nesting of steps and recipes clearer in stats output.

### LINKED ISSUE #434: Better formatting in stats output
Currently, depending on terminal width, the stats output can become very difficult to interpret. This is particularly true with heavily nested recipes. Consider the following:

Image

In the above, the recipe and sub-recipe names have rendered the step names unreadable. I will make an attempt to make this a little more readable by omitting the repeated recipe/sub-recipe names and instead use whitespace/arrows to indicate the relationships between recipes, sub-recipes and steps.

JSKenyon:
Closes #434.

# Before
image

# After
image


o-smirnov (approved):
Lovely, thanks!

PR #441, opened 2025-06-23
Enable console in casatasks and delete extra log

### LINKED ISSUE #440: `casatask` cabs dirty working directories with many logfiles
This issue may belong on the `cult-cargo` repository but I think that the changes to functionality will be here. 

Currently, `casatask` cabs produce casa logs in the current working directory. This often clutters a user's working directory and requires manual clean-up.

I have tried to configure the above behaviour using a `config.py` file but to no avail. The `casatask` cabs seem to generate a stub log no matter what I do. 

My ideal solution is to retain the casa log files (as they may contain valuable information), but move them automatically to the appropriate location in the stimela logs.

JSKenyon:
Closes #440. @o-smirnov I would appreciate your input as while I believe that what I have done works, it is a somewhat bespoke solution.

I think that there is perhaps an argument to be made for having a separate flavour for `casatasks` (even if they are very similar to the `python` flavour) as we will inevitably have to patch in `casatask` specific workarounds in the future and it would be nice to do so without affecting the `python` flavour cabs. My current solution is untested for `casa5` cabs. 

Apologies for all the whitespace changes - I didn't realise until it was too late. 

Edit: The purpose of this PR is to automatically move casa logs to the appropriate place in the stimela log directory.

### DIFF
--- a/stimela/backends/flavours/python_flavours.py
+++ b/stimela/backends/flavours/python_flavours.py
@@ -155,33 +155,24 @@ class PythonCallableFlavour(_CallableFlavour):
         else:
             msg1 = msg2 = msg3 = msg4 = ""
 
-        if py_module == "casatasks" and isinstance(log, logging.Logger):
+        if py_module == "casatasks":
 
-            file_handler = next((h for h in log.handlers if isinstance(h, logging.FileHandler)), None)
-
-            if file_handler is None:
-                log.warning("No file handler has been configured - casa logs will not be moved.")
-            else:
-                log_dir = file_handler.get_logfile_dir()
-
-                logging_prologue = f"""
-import casaconfig
-import shutil
+            logging_prologue = f"""
+from casatasks import casalog
 from pathlib import Path
-casaconfig.get_config()
-casalog_path = Path(casaconfig.config.logfile)
-stimlog_path = Path('{log_dir}/casatask.log')
+casalog_path = Path(casalog.logfile())
+casalog.showconsole(onconsole=True)
 """
 
-                logging_epilogue = f"""
+            logging_epilogue = f"""
 try:
-    shutil.move(casalog_path, stimlog_path)
+    casalog_path.unlink(missing_ok=True)
 except FileNotFoundError:
     pass
 """
 
-                self.pre_command = (self.pre_command or "") + logging_prologue
-                self.post_command = (self.post_command or "") + logging_epilogue
+            self.pre_command = (self.pre_command or "") + logging_prologue
+            self.post_command = (self.post_command or "") + logging_epilogue
 
         code = f"""
 import sys, json, zlib, base64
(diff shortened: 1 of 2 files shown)


o-smirnov:
Will take a look later today, thanks @JSKenyon! But on that subject -- I've been trying to run ``listobs`` as a CASA6 task, and I can't seem to get it to print anything to the console, the obs info seems to go to the log. Have you run up against this with other tasks?

JSKenyon:
> Will take a look later today, thanks @JSKenyon! But on that subject -- I've been trying to run `listobs` as a CASA6 task, and I can't seem to get it to print anything to the console, the obs info seems to go to the log. Have you run up against this with other tasks?

Yeah - it does seem like they are silent. The logger doesn't seem to reach the terminal at all for casatasks from what I can see i.e. simply calling the tasks from `ipython` will only print their return values.

JSKenyon:
Hmm, looking at the source of `casatools.logsink`, I am not sure if it even possible to configure the logger to reach stdout. I am no expert though.

o-smirnov:
Logsink seems to have a ``showconsole()`` method. It also seems to be very proficient at ignoring it...

JSKenyon:
> Logsink seems to have a `showconsole()` method. It also seems to be very proficient at ignoring it...

That seems to work for me if I add it to `logging_prologue` in this PR.

JSKenyon:
And the output does then seem to be captured by Stimela, so perhaps then we don't need to worry about renaming and can just discard the casa log output. That removes the need to bind the log directory. But I am very open to alternatives - this PR doesn't feel as tidy as I would like.

JSKenyon:
@o-smirnov I just pushed a change which I think works and is simple. No worries if you have a more generic solution in mind.

o-smirnov (approved):

o-smirnov:
Looks good, thanks! 

I share your sense of slight untidiness. I wonder if we should not rather be defining the logging stuff [here](https://github.com/caracal-pipeline/cult-cargo/blob/80f8e9cab869b129d2daaaa4f3812ea57fa5cee6/cultcargo/genesis/casa/casa-inputs.yml#L114) than in core stimela?

Perhaps it would be sensible to redefine ``pre_command`` and ``post_command`` as dicts, the values of which are concatenated. That will play better with the usual include/use/merge semantics.

JSKenyon:
> I share your sense of slight untidiness. I wonder if we should not rather be defining the logging stuff [here](https://github.com/caracal-pipeline/cult-cargo/blob/80f8e9cab869b129d2daaaa4f3812ea57fa5cee6/cultcargo/genesis/casa/casa-inputs.yml#L114) than in core stimela?

Absolutely - I think that I had argued myself out of that approach when it was looking like I needed to do something more complicated. I can easily move this there - as it no longer requires input from the calling Python.
 
> Perhaps it would be sensible to redefine `pre_command` and `post_command` as dicts, the values of which are concatenated. That will play better with the usual include/use/merge semantics.

Yeah, that seems sensible. Also gets around the problem of a user-specified pre/post overwriting a built-in one.

JSKenyon:
I am going to close this in favour of a PR on cult-cargo and potentially another PR to address the dictionary issue.

PR #443, opened 2025-06-23
Upgrade pre- and post-command fields to dictionaries.

### LINKED ISSUE #442: Pre- and post- commands should be dictionaries and concatenate commands
Currently, the `pre_command` and `post_command` fields only support a single string. This means that user-specified input will overwrite existing values and potentially break specific cabs. 

The solution is to redefine these fields as dictionaries which provide mappings from arbitrary names to Python code strings.

### Current

```yaml
cabs:
  echo:
    flavour:
      kind: python-code
      pre_command: |
        print("foo")
      post_command: |
        print("baz")
    command: |
      print("bar")

    inputs:
      band:
        dtype: str
```

### Suggested
```yaml
cabs:
  echo:
    flavour:
      kind: python-code
      pre_commands:
        echo-foo: |
          print("foo")
      post_commands:
        echo-baz: |
          print("baz")
    command: |
      print("bar")

    inputs:
      band:
        dtype: str
```

JSKenyon:
Closes #442. This PR replaces `pre_command` and `post_command` with `pre_commands` and `post_commands` which accept dictionaries containing mappings from arbitrary name strings to Python code strings. This allows for multiple pre- and -post commands to be specified with descriptive labels e.g.:

```yaml
cabs:
  echo:
    flavour:
      kind: python-code
      pre_commands:
        echo-1: |
          print("1")
        echo-2: |
          print("2")
      post_commands:
        echo-4: |
          print("4")
        echo-5: |
          print("5")
    command: |
      print("3")
```

I decided to change the name so that it suggests that multiple commands can be added. I have also tried to make the variables consistent across the two Python flavours. I also made minor updates to the docs to reflect these changes.

The only obvious breakage I can think of will be the `casatasks` cabs in `cult-cargo` (I can find no other usage), but I can easily fix that in https://github.com/caracal-pipeline/cult-cargo/pull/137.

### DIFF
--- a/docs/source/reference/cabdefs.rst
+++ b/docs/source/reference/cabdefs.rst
@@ -237,7 +237,7 @@ The following additional options are available for both the ``python`` and ``pyt
 
 * ``flavour.interpreter_command`` determines how the interpreter command line is formed, default is ``"{python} -u"``. Note that this is not subject to Stimela's full {}-substitutions, but does recognize ``{python}``, and inserts ``interpreter_binary`` as set above.
 
-* ``flavour.pre_commands`` adds optional Python code to be executed up front. This can be useful for housekeeping operations, such as disabling warnings, etc. Multiple commands can be specified and will be run in order.::
+* ``flavour.pre_commands`` adds optional Python code to be executed up front. This can be useful for housekeeping operations, such as disabling warnings, etc. Multiple commands can be specified as mappings from arbitrary command names (labels) to code strings and will be executed in order.::
 
     flavour:
         kind: python
--- a/docs/source/reference/cabdefs.rst
+++ b/docs/source/reference/cabdefs.rst
@@ -237,15 +237,16 @@ The following additional options are available for both the ``python`` and ``pyt
 
 * ``flavour.interpreter_command`` determines how the interpreter command line is formed, default is ``"{python} -u"``. Note that this is not subject to Stimela's full {}-substitutions, but does recognize ``{python}``, and inserts ``interpreter_binary`` as set above.
 
-* ``flavour.pre_command`` adds optional Python code to be executed up front. This can be useful for housekeeping operations, such as disabling warnings, etc.::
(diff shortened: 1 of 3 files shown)


docs/source/reference/cabdefs.rst
```diff
@@ -237,15 +237,16 @@ The following additional options are available for both the ``python`` and ``pyt
 
 * ``flavour.interpreter_command`` determines how the interpreter command line is formed, default is ``"{python} -u"``. Note that this is not subject to Stimela's full {}-substitutions, but does recognize ``{python}``, and inserts ``interpreter_binary`` as set above.
 
-* ``flavour.pre_command`` adds optional Python code to be executed up front. This can be useful for housekeeping operations, such as disabling warnings, etc.::
+* ``flavour.pre_commands`` adds optional Python code to be executed up front. This can be useful for housekeeping operations, such as disabling warnings, etc. Multiple commands can be specified and will be run in order.::
```
o-smirnov: Maybe just add a note that the section keys are arbitrary labels and aren't actually part of the command.

stimela/backends/flavours/python_flavours.py
```diff
@@ -154,13 +154,21 @@ def get_arguments(self, cab: Cab, params: Dict[str, Any], subst: Dict[str, Any],
             msg4 = f"""print("## return value is ", _result)"""
         else:
             msg1 = msg2 = msg3 = msg4 = ""
+
+        pre_command_str = ""
+        if self.pre_commands:
+            pre_command_str += "".join(self.pre_commands.values())
```
o-smirnov: Should probably join with "\n", since single-line bits of YAML will not have a trailing newline in the value.

o-smirnov (changes requested):

docs/source/reference/cabdefs.rst
JSKenyon: Added.

stimela/backends/flavours/python_flavours.py
JSKenyon: Done - I think I only had multiline yaml strings in my example. Updated the test accordingly.

o-smirnov (approved):
LGTM, thanks!

PR #446, opened 2025-06-25
Add `-pf/--parameter-file` option

### LINKED ISSUE #445: Add `--parameter-file` option
I have been wondering about the best way to provide default configs for recipes which may need to cater for many different cases. While it is possible to do some of this via `assign_based_on` constructs, this can become quite intricate if there are a sufficient number of orthogonal cases.

For a while I have been wondering if it would be possible to augment `stimela run` command with an additional `-pf/--parameter-file` option which would take one or more `.yml` files and essentially augment the command line with their contents. This is not quite the same as the current recipe file merging which happens.

I have already come up with a prototype solution which seems to work - I will open a PR shortly.

JSKenyon:
Closes #445. This PR adds functionality to specify parameters via  one or more parameter files. This is useful for recipes which have many inputs, or for shipping config variations without mutating recipe defaults. 

For the following recipe (named `recipe.yml`):
```yaml
cabs:
  echo:
    flavour: python-code
    command: |
      for arg in args:
        print(f"{arg}")
      print(f"Found a {extra}!")
    inputs:
      args:
        dtype: List[Any]
      extra:
        dtype: str
        default: gremlin

test-param-file:
  inputs:
    stringy:
      dtype: str
    listy:
      dtype: List[str]
  
  steps:
    echo-inputs:
      cab: echo
      params:
        args:
          - =recipe.stringy
          - =recipe.listy
```
we could use the following parameter file (named `param_file.yml`):
```yaml
stringy: "A string!"
listy:
  - "A"
  - "list"
  - "!"

echo-inputs:
  extra: unicorn
```
and invoke `stimela run recipe.yml test-param-file -pf param_file.yml` to report the existence of a unicorn instead of a gremlin.


o-smirnov (approved):
LGTM!

PR #448, opened 2025-06-27
Rework recipe flow control

### LINKED ISSUE #447: Why does `stimela run -s` require an entry per recipe level?
This is a minor concern but one which I think would aid user friendliness. Currently, if we have a nested recipe e.g. `breifast`, and we want to run a specific step in a sub-recipe, it is necessary to specify `-s` once for each level of nesting e.g. `stimela run main_recipe.yml main-recipe -s subrecipe -s subrecipe.subsubrecipe -s subrecipe.subsubrecipe.step`.

The above seems quite verbose as the first two invocations of `-s` are implied by the key in the final invocation (unless I am missing something). 

I can change this behaviour but I am unsure if there is some subtlety I am missing.

JSKenyon:
While looking at addressing #447, I got myself completely tangled in the recipe flow control logic. This is largely due to the fact that it involves recursion which can become quite difficult to follow. This PR closes #447 but also attempts to simplify the flow control logic. A new class `FlowRestrictor` has been introduced which is initalised with the command line inputs, and can subsequently be queried based on the current recipes `fqname`. This likely needs some more rigorous testing as, while it is a small code change, it is a very different approach which may have edge cases I have not considered. My feeling is that this is an improvement as it means that the state of the `FlowRestrictor` doesn't change between subrecipes which in turn makes it much simpler to figure out what it happening.


JSKenyon:
This still needs more work - in simplifying I broke functionality which didn't have tests. I am now adding tests so that I can go about fixing them.

JSKenyon:
Adding tests has shown me that this functionality isn't working quite as expected on `master` either. In particular, it seems that `always` tags sometimes cause extra steps to run in sub-recipes. The whole thing becomes quite tortuous but I am working on a solution.

JSKenyon:
@o-smirnov what is the intended behaviour when the user specifies a tag on a subrecipe, e.g. `stimela run recipe.yml recipe -t subrecipe.foo`? Does it imply that only that subrecipe should be run, i.e. that all steps besides the step to which the tag is applied should be ignored, or should all steps at or above the level of the tagged step run as normal?

Edit: Current behaviour of `master` seems to be the former i.e. only the tagged steps in the subrecipe and `always` tags get run.

JSKenyon:
This is finally ready for a review @o-smirnov. The number of changes may seem daunting but it is really only the changes to `stimela/kitchen/recipe.py` and `stimela/kitchen/run_state.py` (formerly `stimela/kitchen/utils.py`) which need close scrutiny. The other changes are either minor fixes for issues in the tests, or the new tests themselves. The tests for this functionality may be worth glancing over to ensure that my implementation matches your expectation @o-smirnov.

One thing I will point out in advance is that there is some code duplication at present. This is deliberate. I don't want to start merging functionality into more multipurpose functions until we know that the current code works correctly. It is much easier to fix problems when code can be changed in isolation.

I did profile the new code and it took a few milliseconds for the triply nested test recipe, which is substantially less the the cost of all the Omegaconf merging which is going on. In fact, it iseverything else that dominates the test runtime. If necessary, I could alter the tests to be closer to unit tests i.e. check only the output of `graph_to_constraints` but I think that may miss interactions with the run/skip code.

o-smirnov (approved):
Looks uncontroversial, let's merge this!

PR #453, opened 2025-07-09
fixes tmp binds issue on ilifu

o-smirnov:


o-smirnov:
@JSKenyon please take a look, particularly that I haven't broken local singularity mode.

JSKenyon (approved):
I have not tested this manually but it looks ok to me. I don't think that much has changed for the more common use-cases.

o-smirnov:
There's some new if statements around the local tmp binds, so maybe good to have a local test as well, but I'll do that as I'm running breifast this week.

PR #457, opened 2025-07-10
Fix #456 - `~` characters in logfile substitutions

### LINKED ISSUE #456: Assigning `opts.log.dir` with substitution including `~` results in logs being placed in `_`
Consider the following simple recipe which assigns the name of the output log directory based on the `outdir-path` recipe input:

```yaml
cabs:
  echo:
    command: echo message

recipe:
  assign:
    log.dir: '{recipe.outdir-path}/logs/log-{config.run.datetime}'
  inputs:
    outdir-path:
      dtype: Directory
      required: true
      writable: true
      path_policies:
        mkdir_parent: true
  steps:
    test:
      cab: echo
```

If the user runs this recipe as  `stimela run recipe.yml outdir-path=~/path/to/directory`, some of the logs will end up in `_/path/to/directory`. This happens because the `~` character is replaced with an `_` character [here](https://github.com/caracal-pipeline/stimela/blob/c53af45ee28ae92aeeaabb57dd1c1daeaac1d1da/stimela/stimelogging.py#L328-L329).

The solution should be as simple as calling `os.path.expanduser` before replacing unexpected characters.

JSKenyon:
One-liner which fixes #456. Simply calls `os.path.expanduser` before replacing non-filename characters.

### DIFF
--- a/stimela/stimelogging.py
+++ b/stimela/stimelogging.py
@@ -325,6 +325,8 @@ def update_file_logger(log: logging.Logger, logopts: DictConfig, nesting: int = 
                         log.error(f"bad substitution in log path: {err}")
                     return None
 
+        # Expand path before removing non-filename characters.
+        path = os.path.expanduser(path)
         # substitute non-filename characters for _
         path = re.sub(r'[^a-zA-Z0-9_./-]', '_', path)
 


o-smirnov (approved):
LGTM!

PR #458, opened 2025-07-11
Profiling Redux

JSKenyon:
This is a new version of #308 which was merged and reverted. I have not yet made any changes but I think it needs to be preserved. The reason for the original reversion was performance degradation. I will attempt to fix it at some point.


JSKenyon:
It seems that the use of `interval` in `Process.cpu_percent()` was the cause of the large performance regression. Below is a comparison with master on a nested recipe where mos steps just trigger a 5 second sleep. The impact of the PID aware metrics seems to be about 1 second (~1.5%) over the entire run in this case which I think is probably acceptable.

## Master:
image

## With fix:
image

JSKenyon:
This is now ready for review. It is likely that the `task_stats` module can be rewritten at some point in the future to reduce its reliance on global/module level state. I think that I would consider splitting it out into a separate repository and styling it is a resource monitor that runs in a separate process. This would reduce potential strain on the event loop and potentially allow for higher resolution profiling.

In the interim, this patch should report the CPU and memory usage for stimela in a PID aware fashion while avoiding the massive performance hit @o-smirnov saw after the previous accidental merge. The root cause was an implementation detail of `psutil` - returned `Process` objects are non-unique i.e. we need to manually keep track of the underlying PIDs ourselves to ensure consistent, non-blocking monitoring.

One final thing to note is that I have removed the stimela process itself from the CPU and RAM monitoring as it should always be negligible but can appear to cause strange spikes due to very rapid calls to `psutil`.

stimela/task_stats.py
```diff
@@ -236,6 +238,22 @@ def update_stats(now: datetime, sample: TaskStatsDatum):
...
+def update_children():
+    """Update the module level dictionary mapping child pid to process.
+
+    This is necessary as calling Process.children will return different
+    Process objects each time. These then fail to report CPU stats unless
+    we make them block which has a large impact on performance.
+    """
+    current_children = stimela_process.children(recursive=True)
+    current_pids = {proc.pid for proc in current_children}
+    child_processes.update(
```
o-smirnov, line 250: Isn't that the same as ``child_processes.update(current_children)``, or am I missing something?

o-smirnov: NVM -- I am missing something -- it's a dict not a set.

o-smirnov (dismissed):
LGTM!

o-smirnov:
OK @JSKenyon, if you just resolve the conflict, I'll be happy to merge.

JSKenyon:
Fixed the tiny conflict so should be fine to merge now.

o-smirnov (approved):
LGTM!

PR #461, opened 2025-07-21
Update Progress Display

JSKenyon:
This is currently a work in progress but I am opening the PR to solicit feedback and keep track of what I am doing. The idea is to improve the current progress display so that various performance metrics are much clearer. This includes both task and system information. 

I have also made an attempt to reduce the number of global variables in `task_stats.py` with varying degrees of success. 

Edit: I apologise for including all these whitespace changes. Once these PRs are in I will try and do the same for the entire code base.

Edit: #458 needs to be merged first as this PR is based off of that branch.


o-smirnov:
I like it, but I can still see a use case for the one-liner bar. Can we not have that as a config option?

JSKenyon:
Quite possibly; I am feeling more comfortable with Rich and the task_stats code than I was previously. Out of interest, what use cases do you see for the one-liner bar?

o-smirnov:
The extra real estate for the logs.

JSKenyon:
> The extra real estate for the logs.

Ah ok. I have changed the layout so that it is only costing ~5 lines of logs now:
image

o-smirnov:
> Quite possibly; I am feeling more comfortable with Rich and the task_stats code than I was previously. Out of interest, what use cases do you see for the one-liner bar?

Other two use cases that reoccur to me (how quickly one forgets when not using a mode...): with the slurm backend, there is nothing to show except elapsed time. And with k8s there is barely more than that (just a core count and a pod count, IIRC...) Both would be served better by a one-liner, I feel -- and it should probably be the default for those two backends.

o-smirnov:
On the other hand, with the extra space available in the extended panel, you could show some other useful things. Perhaps system load, utilizing the empty line in the right panel? And perhaps a 5-  or 10-second running average for CPU and RAM use?

o-smirnov:
> On the other hand, with the extra space available in the extended panel, you could show some other useful things. Perhaps system load, utilizing the empty line in the right panel? And perhaps a 5- or 10-second running average for CPU and RAM use?

In the case of RAM, a running max could be even more useful perhaps. Makes it easier to spot short peaks of RAM use that won't necessarily be obvious in the live display.

P.S. Apologies for my tardy reviewing. I plead crazy calendar.

JSKenyon:
> Other two use cases that reoccur to me (how quickly one forgets when not using a mode...): with the slurm backend, there is nothing to show except elapsed time. And with k8s there is barely more than that (just a core count and a pod count, IIRC...) Both would be served better by a one-liner, I feel -- and it should probably be the default for those two backends.

I am trying to get it to the point where I can run it on slurm to check these things. I know that I have definitely broken some of the K8s stuff but am pleading mea culpa until such time as I have a means to test that. By broken, I just mean that the extra stats from K8s aren't being rendered at present. 

> On the other hand, with the extra space available in the extended panel, you could show some other useful things. Perhaps system load, utilizing the empty line in the right panel? And perhaps a 5- or 10-second running average for CPU and RAM use?

Absolutely, I was thinking something similar.

I have been trying to come up with a satisfying way of having different displays for different modes - still iterating.

JSKenyon:
Display now looks as follows @o-smirnov - further feedback welcome.

image

I have also refactored again. There is now a `Display` class which can be configured as `simple` or `fancy`. `simple` produces a single line progress bar similar to the version on `master`. `fancy` produces the multiline progress display above. Additional styles can be added to `Display` as needed, so we could add specific displays for the slurm and k8s backends as needed.

Edit: I still need to decide whether or not we are adding a CLI argument to override the display type, or if we are setting it purely based on the backend settings.

JSKenyon:
@o-smirnov dynamically changing the style turned out to be pretty simple - see the video below. I am not too concerned about the brief period during the change-over when the simple display is in the wrong place (due to the larger display being replaced with the smaller display) - do you agree?

[Screencast from 2025-07-25 09-35-16.webm](https://github.com/user-attachments/assets/d745fabc-4d73-4bc9-b80b-159ec28ef514)

I used the native and singularity backends to check that things were working, but the default will be the simple backend for slurm and kubernetes, and fancy for native and singularity backends. We can add more specialised display styles as needed.

I just need to sort out the manual override now.

JSKenyon:
I have created `display.py` and moved the display related code out of `task_stats.py`. This should make things clearer. I have also added a `remote` style which is used for the `slurm` and `kube` backends. Its is rudimentary at present as I am not sure what we typically want it to show. It should be easy to add more styles/change existing styles now.

I think that this PR can exit draft mode now - all that remains is adding docs/comments.

@Athanaseus if you are up for it, could you possibly try a slurm run with this branch? I just want to be sure that it doesn't explode.

Athanaseus:
Seems to work for me. 

### Using the default:
default


### Using the fancy display:
fancy1

### With no scattering:
fancy0

JSKenyon:
Thanks @Athanaseus; that is super helpful. I think I have something incorrectly configured as I would expect no fancy display on slurm as the local stats are not helpful. Problem for Monday I guess. Could be that the slurm backend isn't detected properly.

JSKenyon:
I have set up a test on ilifu using some advice you posted previously @Athanaseus, thanks! It now uses the correct display on slurm, but not when scattering (the display needs to be correctly set before scattering which it isn't at present). I will fix this shortly.

JSKenyon:
Ok, throwing in the towel for this week. Currently, the display doesn't really do anything useful when scattering a loop. I will need to think about this case (as well as check what happens on master).

Edit: `master` correctly reports status (running steps) but not overall time, so there are definitely things to improve/change for the scattering case.

JSKenyon:
Reverted to draft while I figure out the scatter case.

o-smirnov:
@JSKenyon since this is all swapped into your working memory for now, what do you think of https://github.com/caracal-pipeline/stimela/issues/466#issuecomment-3131612498?

The general solution I was thinking about is:

* Each step in principle has its own set of metrics, depending on its backend, with only elapsed time being common across all steps;
 
* The set of columns being printed should be a logical-OR of the per-step metrics;
 
*  For each step, only print the metrics available and leave the other columns blank.

JSKenyon:
> @JSKenyon since this is all swapped into your working memory for now, what do you think of [#466 (comment)](https://github.com/caracal-pipeline/stimela/issues/466#issuecomment-3131612498)?

I have already added this - `slurm` and `kube` backends should revert to a very simple display which only includes timers, step name and status. I haven't given much thought to the summary - I may defer that to a second PR as this has already ended up being more fiddly than I was anticipating.

> The general solution I was thinking about is:
> 
>     * Each step in principle has its own set of metrics, depending on its backend, with only elapsed time being common across all steps;

Agreed. So technically we may want update methods for each type of runner.

>     * The set of columns being printed should be a logical-OR of the per-step metrics;

Yup, makes sense.

>     * For each step, only print the metrics available and leave the other columns blank.

Blank or `N/A`. I don't object to implementing it, but I think I will make a separate PR once this one is merged to make reviewing less tedious.

JSKenyon:
Slurm does seem to have an `sacct` command can be invoked after a slurm job has run to query various metrics, e.g. ` sacct --format="JobID,JobName,MaxRSS"` will produce something like:
```
JobID           JobName     MaxRSS 
------------ ---------- ---------- 
11268802     sleep.bas+            
11268802.ex+     extern       188K 
11268802.0   sleep.bas+     31780K 
11268803     sleep.bas+            
11268803.ex+     extern       184K 
11268803.0   sleep.bas+     42844K 
11268804     sleep.bas+            
11268804.ex+     extern       200K 
11268804.0   sleep.bas+        76K 
```
Not entirely sure if/how to use the above but perhaps something will occur to me.

o-smirnov:
> I have already added this - `slurm` and `kube` backends should revert to a very simple display which only includes timers, step name and status. I haven't given much thought to the summary - I may defer that to a second PR as this has already ended up being more fiddly than I was anticipating.

The kube backend does collect a couple of useful numbers besides the timer, have you located those and left them in? I'll have a look at what they were called tomorrow...

> Not entirely sure if/how to use the above but perhaps something will occur to me.

Yes, this can be put to use! Let's open a separate enhancement issue for this, but keep it on the mental back burner for now.

JSKenyon:
> The kube backend does collect a couple of useful numbers besides the timer, have you located those and left them in? I'll have a look at what they were called tomorrow...

I have broken the `kube` functionality (well, in that it isn't reporting its extra info at present). This is largely because I cannot test it and doing it by guess-work seemed of dubious value. It wouldn't be too difficult to add back.

If you have a few minutes tomorrow, I could run you through my progress (haha) so far. I am currently stuck on trying to improve the scatter case but things get complicated when multiprocessing is involved.

JSKenyon:
After much chopping and changing, I think that this is now ready for review. There is now a display module which includes a number of styles. The styles include both simple and fancy variants for the various backends. Simple styles are single-line and are approximately equivalent to the current behaviour in `master`. The fancy displays are multi-line and typically include more information with clearer labels. 

I am very open to cosmetic changes - this PR simply splits out and expands the functionality. Changing the appearance of the various displays is not difficult. The one thing I am not completely satisfied with is the way the update methods work but I don't want to change it as part of this PR - it might be quite a large amount of work. There is some heterogeneity in that way the reporting is handled for the various backends. 

One case which is not well-handled in either `master` or this branch is scattering when using the `kube` backend. This is because the backend only starts tracking resources when `run` is invoked (but this is inside the process pool). In principle, it should be possible to have something monitoring the `kube` backend more generally so that we can report the net state of the cluster.

o-smirnov:
Ooops, conflicts now that I've merged the other, older PR @JSKenyon.

Sorry, probably my fault for taking so long with my reviews...

While we're at it, does this play well with https://github.com/caracal-pipeline/stimela/pull/458?

JSKenyon:
It builds on top of it so they can be merged sequentially. Can figure out
the conflicts in the morning if needed.

o-smirnov:
OK, so I'll get busy with the profiling one first then.

JSKenyon:
Ok @o-smirnov, I have resolved the conflicts. If you want any changes to the appearance of the displays, I am open to suggestions.

o-smirnov (changes requested):
LGTM! Thanks for all the hard work, @JSKenyon. 

Some tiny cosmetic requests:

* when the simple display has no stats at the start of the run, better show "--" than "*"

* what is the task "Status" when using the extended display with native mode? I only ever see "N/A" (again, perhaps better "--" than "N/A")

* the read/write ms metric sometimes goes to five digits, which upsets the alignment. Maybe allocate it five digits

JSKenyon:
>     * when the simple display has no stats at the start of the run, better show "--" than "*"

On it - would you like this to be the behaviour everywhere?

>     * what is the task "Status" when using the extended display with native mode? I only ever see "N/A" (again, perhaps better "--" than "N/A")

I inherited it from the old display but I don't have a good handle on when exactly it is meant to be populated. I think that it reports the status of scattered steps e.g. the number of jobs completed.

>     * the read/write ms metric sometimes goes to five digits, which upsets the alignment. Maybe allocate it five digits

Easily done. Will push changes shortly.

o-smirnov:
> 
> On it - would you like this to be the behaviour everywhere?

Yep, "--" for blanks, for consistency.

> I inherited it from the old display but I don't have a good handle on when exactly it is meant to be populated. I think that it reports the status of scattered steps e.g. the number of jobs completed.

Ah right! Well probably best to say "Status: running" then when not scattering (can probably also have a "terminating" status when Ctrl+C is pressed, but that's really just unnecessary frills and flourishes.)

JSKenyon:
Ok, @o-smirnov. I have made the requested changes if you want to take a look. I didn't put in a terminating status - I am happy to do so (it will be mildly fiddly because of the different backends), but can we defer it to a future PR? Really want to minimise the pain when I have to merge master into the linting PR.

o-smirnov:
>  I didn't put in a terminating status - I am happy to do so (it will be mildly fiddly because of the different backends), but can we defer it to a future PR? Really want to minimise the pain when I have to merge master into the linting PR.

By all means, it was just a fancy. As long as we're being fanciful, we can probably report the PID of the process?

o-smirnov (approved):
LGTM!

PR #465, opened 2025-07-28
Fixes #459

### LINKED ISSUE #459: Output put of `File.BASENAME()` should be a string not a `File`
https://github.com/caracal-pipeline/stimela/blob/ddb31c337e1d7fa8b1254da3e4b88b7eefbbe1dc/scabha/basetypes.py#L125-L127

SpheMakh:


scabha/basetypes.py
```diff
@@ -108,30 +107,60 @@ def __repr__(self):
...
+        """
         return File(os.path.abspath(self))
 
     @property
     def DIR(self):
+        """
+        Returns:
+            File: Path to parent directory
+        """
         return File(os.path.dirname(self))
```
o-smirnov, line 143: Should this not be ``Directory``?

o-smirnov (changes requested):

scabha/basetypes.py
sphemakh: It doesn't matter since `Directory` is the as `File`

o-smirnov: True, but it might confuse our future selves so worth making it consistent.

JSKenyon:
@SpheMakh - just checking (as it is relatively new), but have you installed the precommit hooks for ruff?

sphemakh:
No. But it should install with `--with dev` right? 

It didn't in this case because ruff was added after I branched off from master.

JSKenyon:
I think you always need to run `pre-commit install` (once) to configure the hooks. It is opt in functionality for devs but it lets you know if there are problems before the CI gets involved.

sphemakh:
I've been a bit reluctant to switch (especially to `uv`) because it's still very new, but it's done now.

sphemakh:
But change is good, and `uv` is clearly better.

JSKenyon:
`uv` isn't a requirement, but it is quite a bit faster than `pip`. Instructions are here: https://github.com/caracal-pipeline/stimela#installation---developer, for dev installation with both `uv` and `pip`.

sphemakh:
It's not a requirement just like poetry wasn't, but we are clearly moving away from poetry to `uv`; pip is unrelated.

PR #472, opened 2025-08-15
Migrate to uv over poetry.

### LINKED ISSUE #470: Migrate from `poetry` to `uv`
As poetry seems to be a source of constant pain to @o-smirnov (understandably), it is probably time to migrate to a simpler workflow. Will require some changes to the CI but shouldn't be a big deal.

JSKenyon:
Closes #470. `uv` seems to offer a more fully featured and less painful experience than `poetry`. This PR aims to make the switch.

### DIFF
--- a/.github/workflows/python-package.yml
+++ b/.github/workflows/python-package.yml
@@ -29,7 +29,7 @@ jobs:
       with:
         version: "0.8.5"
     - name: Install stimela
-      run: uv sync --extra tests
+      run: uv sync --group tests
     - name: Run stimela to check for successful install
       run: uv run stimela
     - name: Test with pytest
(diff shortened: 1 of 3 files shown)


o-smirnov:
This looks simple! Is ``uv`` provided by gh actions as a standard?

o-smirnov:
Ah nvm, I see you're pulling it in explicitly now...

JSKenyon:
Yeah - it seems to be relatively simple. Not quite sure what to do to test the publish action.

o-smirnov:
Guess we can just push another release candidate and see if it works?

JSKenyon:
I had a comment to this effect, which I have since removed, but I will add linting and other bells and whistles in a subsequent PR. The current linter just spews thousands of lines of complaints which we never read.

o-smirnov (approved):
Looks straightforward!

PR #473, opened 2025-08-18
Add and apply basic `ruff` linting/formatting

JSKenyon:
This PR attempts to tidy up the code in an automated fashion. At present, this is not an automated component of the PR process, but that is the eventual goal.

Currently, I have added the dependency with the default settings and added a max line length of 120 characters - this is quite generous and I don't think it makes the code less readable.

I have not applied other linting fixes yet. Unfortunately, there are a couple of places in the code where `*` imports are used which makes things trickier. 

This is definitely not yet ready to merge, but it is useful to have the PR open to track my progress as run the tests remotely.


JSKenyon:
This is probably a good point to stop. This PR is almost entirely cosmetic. It removes whitespace, formats imports, and enforces a generous 120 character line width. Also checks for obvious evils such as unused imports, unused variables and * imports. These checks are now included in the CI which will mark non-conformant PRs as failing. Developers should install with `--group dev` and then run `pre-commit install`. This will run the linter and formatter on every commit (it is instantaneous), which makes catching problems before the CI simple. We can also consider adding additional rules to the linter if we see a need - I just used the default with a strict 120 character line limit and import sorting.

There is only one place where I had to make a dramatic change that will require a more thorough review. The `configuratt` module was using `*` imports and had two functions calling each other in a circular fashion from separate submodules. I have attempted to resolve the issue by moving these functions to the same file, and getting rid of the `*` imports entirely. This should probably be further improved in a future PR - the fact that the functions call each other in this circular fashion hints that restructuring may be needed.

I will open issues for a couple of other things which I noticed while applying the linting but I will submit the relevant fixes as separate PRs.

Please do not merge this PR until #461 has been merged (if we are going to merge it).

o-smirnov:
> Developers should install with --group dev and then run pre-commit install

OK that just sailed over my head @JSKenyon could you please elaborate? Are these ``uv`` or ``ruff`` commands or what?

> 
> Please do not merge this PR until #461 has been merged (if we are going to merge it).

We certainly are. We just need to get the Ghana TART finished up! Sorry, I've been a terrible reviewer these past few weeks.

JSKenyon:
> OK that just sailed over my head @JSKenyon could you please elaborate? Are these `uv` or `ruff` commands or what?

So I believe that it shouldn't matter whether you use `pip` or `uv` - the syntax should be the same. Examples below.

### `uv` - `uv` managed environment

After cloning the repo, install with `uv sync --group dev`. Then run `uv run pre-commit install` to set up the pre-commit hooks. By default, you should end up with a correctly configured environment in `.venv`. `ruff` can be invoked manually with `uv run ruff check`. Note that this install will not be in editable mode but that can be accomplished by running `uv pip install -e . --group dev`.

### `uv` - `uv pip` managed environment

After cloning the repo, create a virtual environment with `uv venv -p {python_version} path/to/env`. Activate the environment and install with `uv pip install -e . --group dev`. Then run `pre-commit install` inside the environment to set up the pre-commit hooks. `ruff` can be invoked manually with `ruff check`.

### `pip` managed environment

After cloning the repo, create a virtual environment with `virtualenv -p {python_version} path/to/env`. Activate the environment and install with `pip install -e . --group dev`. Then run `pre-commit install` inside the environment to set up the pre-commit hooks.  `ruff` can be invoked manually with `ruff check`.

> We certainly are. We just need to get the Ghana TART finished up! Sorry, I've been a terrible reviewer these past few weeks.

No worries - half the time the comments I leave are to remind myself not to do anything stupid.

o-smirnov:
Great -- can we add this to the README? Add a "For users" section to house the current content (plus maybe a basic pip install command), and a "For developers" section with the above.

JSKenyon:
> Great -- can we add this to the README? Add a "For users" section to house the current content (plus maybe a basic pip install command), and a "For developers" section with the above.

Good idea - I will do so. Will resolve the conflicts at the same time.

JSKenyon:
@o-smirnov I think that this in now ready to merge. I have updated the README and merged the current version of master.

o-smirnov (approved):
Well the code changes are unreviewable... but the yml and toml changes seem reasonable.

JSKenyon:
I hereby commit to fixing anything I broke.

PR #477, opened 2025-08-21
Fixes #475 - remove circular import of main from commands

### LINKED ISSUE #475: CLI creation is circular
I noticed this in passing while working on #473 but didn't want to bloat that PR any further. At present, `stimela` creates a `click.Group` to manage its CLI. This is done in `main.py`. At the bottom of `main.py`, all the subcommands are imported (requiring an import in an unusual place, and none of those commands are actually used in `main.py`). Then, inside the command module files, each command imports the parent CLI group from `main.py`. This works (it probably shouldn't), but it may also be unnecessary - `click.Group` objects can be added to incrementally using `add_command` i.e. it should be possible to import all the subcommands at the top of `main.py`, create the main `click.Group` and then add all the commands, removing the need for the subcommands to import from `main.py`.

JSKenyon:
Fixes #475. Requires #473. Simple change that simplifies the CLI creation to avoid circular imports. Draft until #473 is merged.


o-smirnov (approved):
Great, thanks! I think I followed some pattern from the click tutorial with the ``cli`` object thing, and ended up in a tangle. Glad you've untangled it.

PR #478, opened 2025-08-21
Fixes #476 - successful test runs will purge log files.

### LINKED ISSUE #476: Test log tidying
The test suite invokes `stimela` as a command line application with all of its logging enabled. This is useful but does cause the test directory to gradually fill with log files (which can ultimately irritate IDEs/source control). I would like to add a `conftest.py` which automatically removes the log files after all the tests run successfully. That way, failed test runs will still produce all the logs required for debugging while successful test runs will clear up after themselves.

JSKenyon:
Fixes #476. Requires #473. Draft until #473 is merged.

Running the tests locally gradually generates a gigantic log directory. All log/sundry files created by a stimela test run should now be purged whenever a test run is successful. This is not particularly intelligent code - it doesn't know which tests have passed or failed, it just checks the exit status for the current test run, and if it is zero assumes that it is safe to clean up the generated files.


o-smirnov (approved):
LGTM!

PR #481, opened 2025-08-25
Issue 479 - merge `ImageInfo` dataclasses when using Python backends

### LINKED ISSUE #479: Strange behaviour when using `image` section with no `version` field
Reported by @landmanbester in https://github.com/ratt-ru/breifast/pull/113. For some reason the following:
```yml
cabs:
  dummy:
    image:
      name: python-astro
    flavour: python-code
    command: |
      print("Hello world!")

recipe:
  steps:
    dummy-1:
      cab: dummy
```
ends up pulling an old image (specifically `quay.io-stimela2-python-astro:latest.simg`). The problem does not occur when the `image` field is entirely omitted (or the version explicitly specified). I will dig a little - I am also unsure at this point whether this is a `cult-cargo` or a `stimela` problem.

JSKenyon:
Fixes #479. Depends on #473. 

Adds generic functionality for merging dataclasses. At present, I have set this up such that the output of `scabha.dataclass_utils.merge_dataclass_instances(instance1, instance2)` is a new instance of the same type, but which merges the values of the dataclass fields (`instance2` takes priority over `instance1` except where the field value is `None`). There are other ways to do this - I decided that creating a new instances was the safest as it ensures any `__post_init__` methods will be run. 

I have moved the reconciliation of the default image and user input to the cab `finalize` methods.


o-smirnov (approved):
LGTM!

PR #485, opened 2025-08-27
Add command field to display, fix display when running `build`

### LINKED ISSUE #482: indicate status "building" in progress bar when building images
The old progress bar did have a way of showing when singularity build was being run, but I don't see this in the new progress display, it just shows "image-htc running".

Would be good to have a "image-htc building" in this case.

### LINKED ISSUE #484: Fix `--boring` mode when running `build`
Reported by @o-smirnov. I broke this functionality when changing the display code - I never tested the `build` command (as it is very similar to `run`). This should be a simple fix and can likely include the fix for #482.

JSKenyon:
Fixes #482. Fixes #484. 

Very simple set of changes to prevent `stimela build` from breaking and include the command field in the simple local display.

New appearance:
image

### DIFF
--- a/stimela/commands/run.py
+++ b/stimela/commands/run.py
@@ -254,7 +254,6 @@ def run(parameters: List[str] = [], dump_config: bool = False, dry_run: bool = F
 
     log = logger()
     if not is_logging_boring():
-        display.set_display_style(variant="simple")
         display.enable()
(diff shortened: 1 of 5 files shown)


o-smirnov (approved):
LGTM!

PR #486, opened 2025-08-28
Issue 483 - reduce performance impact of resource monitor on slurm

### LINKED ISSUE #483: Minimise CPU usage of stats/display
Reported by @o-smirnov. Ideally, stimela itself should use minimal resources. It may be running on a login node (e.g. ILIFU), and we should be good neighbours. My recent upgrades to the display and stats have made it slightly greedier (~5% of a core). I will attempt to profile the relevant code and figure out where most of the time is spent. Additionally, it may be sensible to disable/simplify the stats gathering for the slurm backend as, at present, it doesn't report anything meaningful (the login node doesn't do any real work).

JSKenyon:
Fixes #483. This is a work in progress PR which aims to address #483 by dynamically adjusting which resources are being monitored based on the backend. 

At present, I have not looked into the performance impact of the progress/resource display itself. I will look into it once I am satisfied that `--boring` on slurm is sufficiently quiet.


o-smirnov:
Still chewing up 10% for me, at least in build mode.

o-smirnov:
Still eating 2-5%...

o-smirnov:
For comparison, my ``slurm-release-fixes`` branch eats no CPU, despite (uselessly) monitoring the login node.

JSKenyon:
@o-smirnov I think that this is now fairly comparable with your patch release branch. Would you mind taking it for a spin? If you would like it to be even quieter (<<0.6%), just run with `stimela -np` - this will disable the progress display but leave all the pretty logging intact.

The scope crept a bit in the sense that I reworked the resource monitoring to be backend specific. The kube backend needs to be revisited when the cluster is next up - I don't think I have broken it but I would like to move some of its functionality into the new monitoring module. We will need to do so if we ever want to try monitoring the kube backend in a more general sense i.e. from outside the running job. 

I also added `current_backend` and `current_wrapper` properties to `StimelaBackendOptions` in the hopes that it simplifies reasoning about the current backend. Please check that code closely - I may not have understood the resolutions process when multiple backends are specified.

stimela/commands/run.py
```diff
@@ -354,7 +355,7 @@ def run(
     parameter_file: List[str] = [],
 ):
     log = logger()
-    if not is_logging_boring():
+    if not (ctx.parent.params["boring"] or ctx.parent.params["no_progress"]):
```
o-smirnov: Is this a good idea? This checks the command-line arguments of the parent context, correct? Would it not be better to (a) set internal state in the parent and then (b) query that state here? The parent's CLI can change...

stimela/kitchen/recipe.py
```diff
@@ -1344,33 +1345,21 @@ def _run(
...
-                # configured at this point so we cannot track all the pods
-                # running in the scattered loop.
-                if selected_backend == "kube":
-                    display_style = "slurm"
-                elif backend_opts.slurm.enable:
-                    display_style = "slurm"
-                else:
-                    display_style = "local"
+                backend_opts = OmegaConf.to_object(backend_opts)
+                # TODO(JSKenyon): For now, we default to a minimal display for the kube backend
```
o-smirnov: Maybe say "minimal display i.e. "slurm" style" in this comment. I found the display_style = ... if .. else ...  statement a little bewildering otherwise.

stimela/main.py
```diff
@@ -74,6 +74,7 @@ def resolve_command(self, ctx, args):
     help="Reset the configuration cache. First thing to try in case of strange configuration errors.",
 )
 @click.option("--simple", "-S", is_flag=True, help="Use simplified (single-line) displays.")
+@click.option("--no-progress", "-np", is_flag=True, help="Disable the progress display.")
```
o-smirnov: I feel these help strings could be a little clearer. These three options are in increasing order of boringness, correct? They can't be meaningfully combined? Let's try to clarify this in the CLI.

o-smirnov (changes requested):
Some minor queries. Let's try to get this one merged!

stimela/commands/run.py
JSKenyon: Given that argument is an argument of the parent, I would argue that it would make perfect sense for this to break if you change the argument of the parent. Why add extra indirection?

stimela/kitchen/recipe.py
JSKenyon: Noted - will update.

stimela/main.py
JSKenyon: I will need to have a closer look to remind myself of all the locations where these get used. I am not against merging them if it can be dome simply. I worked on this a while ago.

JSKenyon: I added `--no-progress` because `--boring` strips out colour. The main reason for having `--boring` was to make debugging interactively simple (if my memory serves), but the colours don't interfere with that. I will push an update in a moment which makes the options names more consistent i.e. `--simplify-display` and `--disable-display`. I think that keeping them as separate options with slightly more verbose help is better than trying to merge them into one. I am also wary of altering the behaviour of `--boring` as it seems surprisingly involved.

stimela/commands/run.py
o-smirnov: I guess it feels wrong that this function is digging around in the CLI of the parent. This assumes knowledge of the parent's CLI and couples ``run()`` to the parent tighter than seems natural. What if we refatco the code to call run() from some other context?

JSKenyon: I think I am getting confused - the entire reason I went with the above was because the parent `stimela` CLI code is always invoked. I was working around the fact that `--boring` was an input to the `stimela` command and not the `stimela run` command. Naturally, if we moved `--boring` (and the other progress display options) to the specific commands (build/run), there would be no need to request the parameter from the parent. I just figured that this was a more programmatic way of accomplishing things without communicating via internal state.

o-smirnov: It still sits a little uneasy with me, but it's a pretty superficial piece of logic anyway, so let's not hold up this PR anymore. We can always revise it later if I come up with more persuasive objections.

o-smirnov (approved):
LGTM!

PR #505, opened 2025-10-15
Issue 502

SpheMakh:
Including updates to the README


.github/workflows/python-package.yml
```diff
@@ -36,5 +36,5 @@ jobs:
       run: uv sync --group tests
     - name: Run stimela to check for successful install
       run: uv run stimela --help
-    - name: Test with pytest
-      run: uv run pytest tests
+    - name: Test with pytestuv run pytest tests
+      run:
```
JSKenyon: This will have disabled the tests.

pyproject.toml
```diff
@@ -37,6 +37,7 @@ docs = [
     "Sphinx>=5.3.0,<6",
     "sphinx-copybutton>=0.5.0,<0.6",
     "furo>=2022.9.15,<2023",
+    "docutils"
```
JSKenyon: I would suggest including a rough minimum version.

stimela/backends/runner.py
```diff
@@ -52,6 +54,9 @@ def validate_backend_settings(backend_opts: Dict[str, Any], log: logging.Logger)
     selected = backend_opts.select or ["singularity", "native"]
     # select containerization engine, if any
     for name in selected:
+        # container tech cannot be used if cab.image has not been set
+        if name in ["singularity"] and not getattr(cab, "image", None):
+            continue
```
JSKenyon: If `selected` only includes `singularity`, but no cab was defined, will this loop behave as expected? Is there are reason for using `in` over `==`?

JSKenyon (dismissed):
Took a quick look; just an accidental change to revert and a few questions. @o-smirnov should also check that this won't have unexpected consequences.

stimela/backends/runner.py
o-smirnov: This ought to say ``in ["singularity", "kube"]``, as both of these absolutely require ``cab.image``.

If selected is only singularity, and there's no cab.image, the loop will fall through to raise a BackendError. Which is as expected I think -- but the error message is going to be a little misleading, since it's not that that the backend is not available -- it's that the cab doesn't support it without an image.

stimela/backends/runner.py
```diff
@@ -39,7 +39,9 @@ def build(self, cab: "stimela.kitchen.cab.Cab", log: logging.Logger, rebuild=Fal
             return self.backend.build(cab, backend=self.opts, log=log, rebuild=rebuild, wrapper=self.wrapper)
 
 
-def validate_backend_settings(backend_opts: Dict[str, Any], log: logging.Logger) -> BackendRunner:
+def validate_backend_settings(
+    backend_opts: Dict[str, Any], log: logging.Logger, cab: "stimela.kitchen.cab.Cab" = None
+) -> BackendRunner:
```
o-smirnov, line 49: You've added a cab argument, but you haven't changed the validate_backend_settings() invocations to actually pass it -- or am I missing something?

o-smirnov (changes requested):
See comments -- I'm missing where cab is passed in?

stimela/backends/runner.py
sphemakh: > If selected only includes singularity, but no cab was defined, will this loop behave as expected?
It will behave the same way it would've if only singularity was selected but wasn't installed. So this addition does break the original behaviour of this loop.

> Is there are reason for using in over ==?
Yes, we should add runners for other containerization tools at some point. When we do, you just add it to the list.

stimela/backends/runner.py
sphemakh: Oopsie, I didn't commit that file. Adding, I'll also add a test for this.

.github/workflows/python-package.yml
sphemakh: I have no idea how this was changed. Reverting back

stimela/backends/runner.py
o-smirnov: > but the error message is going to be a little misleading, since it's not that that the backend is not available -- it's that the cab doesn't support it without an image.

Any ideas for a better error message? "Selected backend(s) not available, or no cab image configured"?

stimela/kitchen/step.py
```diff
@@ -437,7 +434,7 @@ def run(
                 log_rich_payload(self.log, "current backend settings are", opts_yaml, syntax="yaml")
             backend_opts = OmegaConf.merge(StimelaBackendSchema, backend_opts)
             backend_opts = OmegaConf.to_object(backend_opts)
-            backend_runner = runner.validate_backend_settings(backend_opts, log=self.log)
+            backend_runner = runner.validate_backend_settings(backend_opts, self.cargo, log=self.log)
```
o-smirnov: @SpheMakh have you tested this with nested recipes? I have a sneaky suspicion this might fail if the cargo is a recipe, and the backend is singularity only.

I think the correct thing to do to pass ``self.cargo if isinstance(self.cargo, Cab) else None``.

I think that should be done when invoking in finalize() and run(). No need in build() since there the if statement checks for a Cab anyway.

stimela/backends/runner.py
```diff
@@ -60,7 +75,10 @@ def validate_backend_settings(backend_opts: Dict[str, Any], log: logging.Logger)
                 backend_name = name
                 break
     else:
-        raise BackendError(f"selected backends ({', '.join(selected)}) not available")
+        raise BackendError(
+            f"selected backends ({', '.join(selected)}) not available, "
+            f"or the cab '{cab.name}' does not specify a container image"
```
o-smirnov: This will fail if cab is None.

stimela/backends/runner.py
o-smirnov: Should be Optional[Cab] because of, see comment below (if the cargo is a recipe).

o-smirnov (changes requested):

stimela/backends/runner.py
sphemakh: Maybe we should add a `requires_cab_image` flag to the backend schema. Then have an `if-else` clause when raising an exception (which can also be more informative (missing backend or missing image)

stimela/kitchen/step.py
sphemakh: I was doing this check in `runner.validate_backend_settings()` but makes more sense here.

stimela/backends/runner.py
o-smirnov: But if we add this as a field to the backend schema, it becomes something that can be settable by the user, do we really want that? Isn't it a static class-level property rather?

I think to make the error more informative, we should rather accumulate a list of "reasons" why we couldn't use any of the selected backends. Something along these lines:

```python
   excuses = {}
   for name in selected:
        opts = getattr(backend_opts, name, None)
        # explicitly disabled
        if opts and not opts.enabled:
          excuses[name] = "not enabled"
          continue
        backend = get_backend(name, opts)
        if backend is None:   # will be the case if e.g. singularity or k8s not installed on system
          excuses[name] = get_backend_status(name)
          continue
        # container tech cannot be used if cab.image has not been set
        if name in ["singularity", "kube"] and isinstance(cab, Cab) and cab.image is None:
          excuses[name] = "cab does not define an image"
          continue
        # success
        backend_name = name
        break
   else:
        reason = ", ".join([f"{name}: {why}" for name, why in excuses.items()])
        raise BackendError(f"unable to select a backend ({reason})")
```

sphemakh: There is no base `BackendSchema` that is inherited by the different backend schemas, instead there are defining attributes in the backend modules (a few common ones too, which may require a re-think later on) but that's where I added a `requires_container_image` attribute which can be checked when validating `backend.select`

sphemakh: We definitely need to consolidate these backend schemas/modules because it's quite confusing  and has a lot of duplication.

stimela/backends/runner.py
```diff
@@ -39,28 +40,49 @@ def build(self, cab: "stimela.kitchen.cab.Cab", log: logging.Logger, rebuild=Fal
...
+                    if isinstance(cab, Cab) and not cab.image:
+                        continue
                 break
     else:
-        raise BackendError(f"selected backends ({', '.join(selected)}) not available")
+        selected_string = ", ".join(selected)
+        # backend is available but cab.image is not set when cab is a Cab.
+        if backend_name:
+            raise BackendError(
+                f"Selected backdend(s) '{selected_string}' require a container image but"
```
o-smirnov: Typo: backdend(s). Didn't you like my pattern for accumulating excuses?

o-smirnov: I think this logic for reporting the error will not always work -- you reset backend_name back to None at the start of the loop. So let's say backend.select is ``singularity,native``, and the native backend is disabled -- the error message will just say "not available".

o-smirnov (changes requested):
Don't think the error reporting will be right, see comment...

stimela/backends/runner.py
sphemakh: If singularity is available, then it picks it, if it's not then it fails with 'backend not available' which is what it should do right?

sphemakh: > Didn't you like my pattern for accumulating excuses?

Nope... A `requires_container_image` attribute makes more sense; it's similar to the `is_remote` attribute and belongs in the backend module/definition.

o-smirnov: > If singularity is available, then it picks it, if it's not then it fails with 'backend not available' which is what it should do right?

I think it will just report that native is not available, and won't mention the missing image.

o-smirnov: > 
> Nope... A `requires_container_image` attribute makes more sense; it's similar to the `is_remote` attribute and belongs in the backend module/definition.

I meant a different pattern, very pretty, which I typed up and didn't post. Let me re-create it quickly...

o-smirnov: OK how about this:

```python
    excuses = {}
    for name in selected:
        # check that backend has not been disabled
        opts = getattr(backend_opts, name, None)
        if opts and not opts.enable:
          excuses[name] = "disabled"
          continue
        backend = get_backend(name, opts)
        if backend is None:
          excuses[name] = get_backend_status(name)  # will tell what's wrong
          continue
        if getattr(backend, "requires_container_image", False) and isinstance(cab, Cab) and not cab.image:
          excuses[name] = "container image not specified by cab"
          continue
        # this one wins
        backend_name = name
        break
    else:
        reason = "; ".join(f"{name}: {excuse}" for name, excuse in excuses.items()})
        raise BackendError(f"unable to select a backend ({reason})")
```

sphemakh: The test passes, and it fails with the right message for the case you raise. See 
https://github.com/caracal-pipeline/stimela/blob/5993f679c114bb0c573a8dc64e30b8c596b93971/tests/stimela_tests/test_backend_validation_singularity_native.py#L140-L151

and the two tests below this one

o-smirnov: But I think in this case it should give two reasons. Because the two backends are not available for different reasons. Hence my "excuses" suggestion above.

sphemakh: Noted, I'll accumulate the reasons for each backend failure and log each in the error in the error message.

sphemakh: > OK how about this:
> 
> ```python
>     excuses = {}
>     for name in selected:
>         # check that backend has not been disabled
>         opts = getattr(backend_opts, name, None)
>         if opts and not opts.enable:
>           excuses[name] = "disabled"
>           continue
>         backend = get_backend(name, opts)
>         if backend is None:
>           excuses[name] = get_backend_status(name)  # will tell what's wrong
>           continue
>         if getattr(backend, "requires_container_image", False) and isinstance(cab, Cab) and not cab.image:
>           excuses[name] = "container image not specified by cab"
>           continue
>         # this one wins
>         backend_name = name
>         break
>     else:
>         reason = "; ".join(f"{name}: {excuse}" for name, excuse in excuses.items()})
>         raise BackendError(f"unable to select a backend ({reason})")
> ```

This works, I'll add it.

o-smirnov: It works? I'm pleasantly surprised because I just typed it up on the fly :rofl:

o-smirnov (approved):
LGTM!

stimela/backends/runner.py
sphemakh: Welcome to the *Shirai Ryu*

PR #509, opened 2025-10-29
Fix #503 - Only log messages on state change

### LINKED ISSUE #503: Continuous log message from stimela with k8s backend: " container state is 'running'"
Using k8s backend with Stimela is outputting a message saying " container state is 'running'"  every 100 milliseconds or less. 

Makes it hard to scroll through the logs and understand what happened. 
Attaching an example of WSclean log file.

[log-MKtest.clean-image.txt](https://github.com/user-attachments/files/22920781/log-MKtest.clean-image.txt)

JSKenyon:
Fixes #503. Simple fix which prevents the k8s backend from spamming messages during pod termination.


o-smirnov (approved):
Very civilized fix, thanks @JSKenyon!

PR #515, opened 2026-02-23
Fix #514 - remove defunct referece to `missing_parent`

### LINKED ISSUE #514: `Key 'missing_parent' not in 'FailRecord'` - optionals trigger edge case in cache
Discovered by accident while assisting a user. The code in question is triggered by optional includes when checking for cache invalidation. If an optional file has appeared between runs, the cache must be invalidated. However, it is currently triggering old code which refers to a non-existent attribute, causing a crash. Fix incoming.

JSKenyon:
Fixes #514. Simple change to remove old code and ensure that optionals only trigger cache invalidation when they appear.

### DIFF
--- a/scabha/configuratt/deps.py
+++ b/scabha/configuratt/deps.py
@@ -176,9 +176,8 @@ class ConfigDependencies(object):
                         return True
                 except ImportError:
                     pass
-            elif not dep.missing_parent:
-                if os.path.exists(filename):
-                    return True
+            elif os.path.exists(filename):
+                return True
         return False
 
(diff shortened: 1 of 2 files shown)


o-smirnov (dismissed):
LGTM!

o-smirnov (commented):
Haha I always had a deep mistrust of ruff and other iedologically driven formatters. :P

JSKenyon:
I suspect that a newer version has made a minor change which caused those lines to change. All good - just had to run `ruff format` locally.

o-smirnov (approved):

PR #519, opened 2026-03-04
fixes #518

### LINKED ISSUE #518: Add ability to issue conditional warnings and errors to poor users
It would be nice to issue warnings (and errors) when certain conditions are met. Useful to check that parameters are consistent, that senseless runtime coditions haven't arisen, etc.

This should be definable at cab and recipe level, but also at step level.

The syntax would be an optional field at cab/step level, let's call it ``sanity_checks`` but open to alternative names, defined as a sequence of:

```yml
sanity_checks:
  - condition: =VALID(recipe.ms)
    message: MS must be valid
    severity: abort 
  - condition: =current.foo >= 0
    message: negative foo!
    severity: warning
```

Issue warning (or abort entirely). Similar to thoughts in #371 --

* Perhaps print only the first time it arises?

* Collect all warnings into a list, and print again at the end of the run

See also here: https://github.com/caracal-pipeline/stimela/issues/438#issuecomment-2999345508.

@JSKenyon @SpheMakh @landmanbester thoughts?

o-smirnov:


o-smirnov:
@JSKenyon please have a look when you've got a chance. See ``test_notify.yml`` for current usage.

stimela/kitchen/step.py
```diff
@@ -120,6 +123,14 @@ def __post_init__(self):
         else:
             # otherwise, self._skip stays at None, and will be re-evaluated at runtime
             self._skip = None
+        # check that preamble/epilogue entries are valid (each must be a list of expressions)
+        from omegaconf import ListConfig
```
JSKenyon: Does this need to be a runtime import?

JSKenyon (dismissed):
Looks good to me - just one tiny question about an import. Was this mostly written by hand or by Claude? Still trying to get a feel for it.

stimela/kitchen/step.py
o-smirnov: Don't see why -- fixed!

o-smirnov:
99% Claude. Not sure why it stuck that import there, but I fixed it. Could you please review again @JSKenyon?

JSKenyon (approved):
Looks good to me!

PR #525, opened 2026-03-27
fixes #517

### LINKED ISSUE #517: Add a DEFAULT-RECIPE key at top level?
The command line e.g.

``stimela run breifast.recipes::spectron.yml tron obs=L2-pfb``

has a certain redundancy to it, especially when we tend to implement a "one app - one YAML doc" pattern. Propose adding a ``DEFAULT-RECIPE: tron`` key at the end of the recipe file, meant for application-level recipes like these. The logic is simple:

* if a name is given on the command like, run/doc that recipe or cab (current behaviour)

* if not given, check for a DEFAULT-RECIPE key, and run/doc that (new behaviour)

* if neither of the above are given, display "ERROR: found multiple recipes or cabs, please specify one on the command line" (current behaviour)

@JSKenyon @SpheMakh @landmanbester thoughts?

o-smirnov:


JSKenyon (dismissed):
Works for me - my only concern is whether DEFAULT is perhaps a little vague? I thought the original proposal was for a DEFAULT_RECIPE field.

o-smirnov:
> Works for me - my only concern is whether DEFAULT is perhaps a little vague? I thought the original proposal was for a DEFAULT_RECIPE field.

It was, but the I realized it can also define a default cab. I considered DEFAULT_RUNNABLE, but why complicate things?

JSKenyon:
> > Works for me - my only concern is whether DEFAULT is perhaps a little vague? I thought the original proposal was for a DEFAULT_RECIPE field.
> 
> It was, but the I realized it can also define a default cab. I considered DEFAULT_RUNNABLE, but why complicate things?

Explicit is better than implicit but this isn't confusing enough for me to go to war over. :-)

o-smirnov:
Fixed Copilot's comments -- @JSKenyon could you please re-review?

JSKenyon (approved):
Looks good to me!

PR #529, opened 2026-04-08
fix formula-to-formula evaluation marking as unresolved during prevalidation

### LINKED ISSUE #527: Substitution regression after #526
PR #526 seems to have introduced a regression in the substitution behaviour. The following simple example:
```yaml
cabs:
  echo:
    command: echo
    inputs:
      msg:
        dtype: str

outer:
  inputs:
    ncpu:
      dtype: int
      default: 1

  steps:
    inner-1:
      recipe: inner
      params:
        ncpu: =recipe.ncpu

inner:
  inputs:
    ncpu:
      dtype: int

  steps:
    echo-1:
      cab: echo
      params:
        msg: =recipe.ncpu
```
will fail with `ncpu = =recipe.ncpu: value is not a valid integer`. Note that the problem doesn't occur when there is no nested recipe call i.e. if the `echo` cab is called directly from `outer`, rather than via `inner`.

### LINKED ISSUE #528: self-referential substitutions should cause an error
The changes that Copilot nagged me into as part of https://github.com/caracal-pipeline/stimela/pull/526 have broken breifast. An unintended change has revealed a prior bug.

Fundamentally, what happens is that ``ncpu: =recipe.ncpu`` now gets passed into a sub-recipe during *prevalidation*, when things are still allowed to be unresolved (whereas before I suspect it got stripped out on account of being unset). This now means we have ``recipe.ncpu: =recipe.ncpu`` inside the subrecipe. The ``=recipe.ncpu`` substitution happily evaluates to the string ``"=recipe.ncpu"`` without causing an error -- arguably it should! -- but down the line we attempt to convert it into an integer and fall over in disgrace.

Solutions:

* [ ] revert to old behaviour i.e. strip out unset parameters during prevalidation, as we used to before #526 got merged.

* [x] cry in disgust at recursive substitutions like the above?

o-smirnov:
When a sub-recipe receives a formula param like ncpu: =recipe.ncpu, the inner recipe's namespace update replaces subst.recipe with the raw step params (containing the formula string). The evaluator then resolves recipe.ncpu to the formula string itself, which passes through unchanged and fails type validation. Now, formulas that evaluate to formula strings are correctly marked as Unresolved and deferred to runtime.

Fixes #527 and parts of #528

### DIFF
--- a/scabha/evaluator.py
+++ b/scabha/evaluator.py
@@ -882,7 +882,7 @@ class Evaluator(object):
                     except (AttributeError, SubstitutionError, ParserError, FormulaError) as err:
                         if raise_substitution_errors:
                             raise
-                        new_value = Unresolved(errors=[str(err)])
+                        new_value = Unresolved(errors=[err])
                     # if a formula evaluated to a formula string (e.g. because the namespace
                     # contains the raw formula), treat as unresolved
                     if (
@@ -893,7 +893,14 @@ class Evaluator(object):
                         and new_value.startswith("=")
                         and not new_value.startswith("==")
                     ):
-                        new_value = Unresolved(errors=[f"formula '{value}' could not be fully resolved"])
+                        new_value = Unresolved(
+                            errors=[
+                                FormulaError(
+                                    f"{'.'.join(self.location + sublocation)}: "
+                                    f"formula '{value}' could not be fully resolved"
+                                )
+                            ]
+                        )
                     if verbose:
                         print(f"{name}: {value} -> {new_value}")
                     # UNSET return means delete or revert to default
--- a/scabha/evaluator.py
+++ b/scabha/evaluator.py
@@ -883,6 +883,17 @@ class Evaluator(object):
                         if raise_substitution_errors:
                             raise
                         new_value = Unresolved(errors=[str(err)])
+                    # if a formula evaluated to a formula string (e.g. because the namespace
+                    # contains the raw formula), treat as unresolved
+                    if (
+                        not raise_substitution_errors
+                        and type(new_value) is str
+                        and value.startswith("=")
+                        and not value.startswith("==")
+                        and new_value.startswith("=")
+                        and not new_value.startswith("==")
+                    ):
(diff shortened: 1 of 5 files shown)


o-smirnov:
OK @JSKenyon ready for you.

scabha/evaluator.py
```diff
@@ -882,7 +882,25 @@ def evaluate_dict(
...
+                        and not new_value.startswith("==")
+                    ):
+                        new_value = Unresolved(
+                            errors=[
+                                FormulaError(
+                                    f"{'.'.join(self.location + sublocation + [name])}: "
+                                    f"formula '{value}' could not be fully resolved"
+                                )
+                            ]
+                        )
```
JSKenyon: Am I correct in saying that this checks whether `value` looks like a formula string and then wraps it as an Unresolved? When does resolution happen?

JSKenyon (dismissed):
One minor comment, more for my education than in opposition. Changes look simple and the test will hopefully protect against this problem in future.

scabha/evaluator.py
o-smirnov: Correct. It doesn't happen -- think of it as "unresolvable" rather than "unresolved". The Unresolved value gets picked up upstream and reported back to the user eventually.

Thanks for making me eyeball that though, as I've spotted an issue with the fix. Better fix incoming!

o-smirnov:
Sorry @JSKenyon can you please re-review!

tests/stimela_tests/test_recipe.py
```diff
@@ -122,6 +122,33 @@ def test_test_loop_recipe():
     assert retcode == 0
 
 
+def test_issue527_1():
+    """Test that circular references are ignored in prevalidation"""
```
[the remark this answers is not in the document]
o-smirnov: Nope. The passthrough created the circularity, but it's the circularity that is being tested here.

JSKenyon (approved):
Looks sensible and I am all for more tests!

PR #541, opened 2026-04-28
The scabha divorce

### LINKED ISSUE #500: The scabha independence
@caracal-pipeline/stimelators 

`scabha` was moved into `stimela` for rapid development in the early stages of `stimela 2.0`. But now that things are stable, we've decided (with @o-smirnov ) it's time time to set it free again. I've copied the  `scabha` files in the *master* branch to the [scabha repo](https://github.com/caracal-pipeline/scabha), and will delete them via an incoming PR. 

This change will not affect users; `scabha` will just become another dependency.

For the devs, let's please merge all `scabha` changes to the *master* branch in the next two months.

scabha

o-smirnov:


stimela/kitchen/step.py
```diff
@@ -318,7 +318,7 @@ def prevalidate(self, subst: SubstitutionNS, root=False, backend=None):
             self.cargo.apply_dynamic_schemas(params, subst)
             # will prevalidate again below based on these updated schemas
         # validate cab or recipe
-        params = self.validated_params = self.cargo.prevalidate(params, subst, root=root)
+        params = self.validated_params = self.cargo.prevalidate(params, subst)
```
JSKenyon: Has `root` been removed intentionally?

JSKenyon (dismissed):
The changes are relatively minimal at the end of the day. I don't see anything problematic; I had one question about a small change.

stimela/kitchen/step.py
o-smirnov: I think this is a mistake. @SpheMakh please comment.

sphemakh: I think removed it because it's not used in the `Cargo.prevalidate` same `backend` actually. But I'll add back and we can how to deal with these cases later.

o-smirnov: Ah but remember, Cab and Recipe are subclassed from Cargo, and these classes **do** use the ``root`` and ``backend`` arguments. The call to ``self.cargo.prevalidate()`` call actually goes to their methods. So these arguments need to be passed through.

That said, I see there was a bug in Step all along -- ``backend`` was not actually being passed to ``cargo.prevalidate()``. It's meant to be passed.

Affected functionality is not covered by current tests, so it went unnoticed:

* catching backend specification errors early (in prevalidation)

* the ``root`` substitution namespace, which was arguably a dumb idea anyway (and is not used in anger thankfully)

Athanaseus (approved):

PR #555, opened 2026-05-19
Relax python version.

### LINKED ISSUE #554: Changing to uv created a subtle restriction for testing with python 3.14
Developing tart_cargo that works fine with python 3.14. Prior to conversion to UV (at a guess) upstream dependency on stimela worked fine with my pyproject.toml having

     requires-python = "<4.0,>=3.9"

tart_cargo can depend on stimela ^2.0.1.1, but not more recent versions as stimela now requires python <3.14 in pyproject.toml

Is there a recommended way to depend on latest stimela and allow 3.14 so we can keep testing on current python versions?

* Python 3.9 End of Life / Final Security Release (3.9.25): October 31, 2025.  Because this version reached its EOL in late 2025, it no longer receives security patches or updates. It is highly recommended to upgrade to a supported version, such as [Python 3.12 or 3.13](https://www.python.org/downloads/). 
* Python 3.10 was first released on October 4, 2021. It is currently in its security-only phase, meaning it only receives source-only security updates and no longer receives binary installers or bug fixes".

JSKenyon:
I believe that (with a very, very quick look) that we can fix #554 by relaxing the python version. Let's see if the CI passes the tests.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -6,7 +6,7 @@ authors = [
     { name = "Oleg Smirnov and RATT", email = "osmirnov@gmail.com" },
     { name = "Sphesihle Makhathini", email = "sphemakh@gmail.com" },
 ]
-requires-python = ">=3.9,<3.14"
+requires-python = ">=3.9,<=3.14"
 readme = "README.rst"
 license = "GPL-2.0-only"
 license-files = ["LICENSE"]


o-smirnov (approved):
Happy to approve, why did we discriminate against 3.14 anyway?

PR #560, opened 2026-05-21
Avoid counting shared memory segments multiple times

adevress:
## Summary
Avoid to count the shared memory segments multiple times in the native reporting for Stimela.

## Changes
This fix is keeping track of the inodes associated with shmem segment during the counting.
This should fork for both /dev/shmem and SYSV- shmem.
memfd might still causes a duplication

## Tests
Has been tested in isolation.
Should be tested with the faulty process before merge

### DIFF
--- a/stimela/monitoring/local.py
+++ b/stimela/monitoring/local.py
@@ -23,11 +23,11 @@ def update_children():
         del _child_processes[pid]
 
 
-def get_shared_memory_usage(pid: int, shm_segments_info: Dict[str, int]):
-    """
+def get_shared_memory_usage(pid: int, shm_segments_info: Dict[int, bool]):
+    """Get shared memory usage of a process without duplication.
+
     Getting shared memory usage of multiple processes without duplication
     is tricky under Linux. This is an attempt to fix that.
-
 
     Args:
         pid: pid of the process to check
--- a/stimela/monitoring/local.py
+++ b/stimela/monitoring/local.py
@@ -1,4 +1,5 @@
 from dataclasses import dataclass
+from typing import Dict
 
 import psutil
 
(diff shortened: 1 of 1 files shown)


JSKenyon:
@adevress I am taking a look at this at the moment. Can you remember the specific example on which you saw the problem? I am trying to come up with a simple example to test this but haven't managed to trip up the master branch yet. I definitely recall that this is a problem but now I cannot remember the exact step/application in which it occurs.

JSKenyon:
Never mind - I wasn't touching all the memory in the child processes to the reporting issue wasn't triggered. I have a stimela recipe which reproduces and suggests that this fix works as intended. 

```yaml
## Stimela recipe: place a 4 GB array in POSIX shared memory and have a pool
## of worker processes attach to it. The workers do no work — each simply
## attaches to the existing shared-memory block, sleeps, and returns. This is
## useful for exercising/observing shared-memory behaviour (e.g. RSS vs. shared
## accounting) without any real computation.

cabs:
  shared-mem-attach:
    info: "Allocate a large shared-memory array and attach to it from N processes."
    flavour:
      kind: python-code
      # Inputs are exposed as plain Python variables inside the snippet.
      input_vars: true
      output_vars: true
    inputs:
      size_gb:
        dtype: float
        default: 4.0
        info: "Size of the shared-memory array in gibibytes."
      n_procs:
        dtype: int
        default: 4
        info: "Number of worker processes that attach to the shared array."
      sleep_seconds:
        dtype: float
        default: 15.0
        info: "Time each worker sleeps before returning, in seconds."
    outputs:
      block_name:
        dtype: str
        info: "Name of the shared-memory block that was created."
    command: |
      import os
      import time
      import multiprocessing as mp
      from concurrent.futures import ProcessPoolExecutor
      from multiprocessing import shared_memory

      # Allocate the shared-memory block. We size it in bytes from the requested
      # gibibytes and back it with a numpy array view so the memory is real
      # (touched) rather than lazily reserved.
      import numpy as np

      n_bytes = int(size_gb * (1024 ** 3))
      shm = shared_memory.SharedMemory(create=True, size=n_bytes)
      block_name = shm.name

      # Build an array view over the block and fault every page in so the full
      # 4 GB is genuinely resident in shared memory.
      array = np.ndarray((n_bytes,), dtype=np.uint8, buffer=shm.buf)
      array[:] = 1
      print(f"Created shared-memory block {block_name!r} of {n_bytes} bytes")

      def _worker(name: str, nbytes: int, sleep_for: float) -> int:
          """Attach to an existing shared-memory block, sleep, then detach.

          Args:
              name: Name of the shared-memory block to attach to.
              nbytes: Size of the block in bytes.
              sleep_for: Seconds to sleep while attached.

          Returns:
              The PID of the worker process that attached.
          """
          child = shared_memory.SharedMemory(name=name)
          view = np.ndarray((nbytes,), dtype=np.uint8, buffer=child.buf)
          pid = os.getpid()
          # Read EVERY page of the block so the full array becomes resident in
          # THIS process's RSS. Because the block is MAP_SHARED, the physical
          # pages are the very same ones the parent and sibling workers use --
          # total physical RAM stays ~size_gb -- but each process's RSS now
          # independently reports the whole array. Stimela sums
          # memory_info().rss across the process tree, so N workers each
          # faulting in the block makes it report ~N * size_gb of "used" RAM.
          checksum = int(view.sum())
          print(f"[pid {pid}] attached, faulted in {nbytes} bytes "
                f"(checksum={checksum}), sleeping {sleep_for}s")
          time.sleep(sleep_for)
          child.close()
          print(f"[pid {pid}] detached")
          return pid

      # Run the workers via a ProcessPoolExecutor. We pin it to the "fork" start
      # method so the forked workers inherit _worker from this snippet's globals
      # (an inline snippet has no importable module for "spawn" to re-import).
      ctx = mp.get_context("fork")
      with ProcessPoolExecutor(max_workers=n_procs, mp_context=ctx) as executor:
          futures = [
              executor.submit(_worker, block_name, n_bytes, sleep_seconds)
              for _ in range(n_procs)
          ]
          for future in futures:
              future.result()

      # Clean up: detach the parent view and release the block.
      del array
      shm.close()
      shm.unlink()
      print(f"Unlinked shared-memory block {block_name!r}")

shared-mem-demo:
  name: "shared-memory attach demo"
  info: "Create a 4 GB shared array and attach to it from 4 idle worker processes."
  steps:
    attach:
      cab: shared-mem-attach
      params:
        size_gb: 4.0
        n_procs: 4
        sleep_seconds: 15.0
```

JSKenyon:
If you have a moment and are interested at some point @adevress, there is an alternative PR https://github.com/caracal-pipeline/stimela/pull/562 up for this. I would appreciate your input on which one is likely to be more robust.

PR #587, opened 2026-06-25
Allow '+' and '@' characters in log file paths

### LINKED ISSUE #513: `+` character in log name
I am not sure if this qualifies as a bug but it did confuse me for a while. Targets are often referred to by their J2000 coordinates/name. This regularly includes a `+` character. If the name of the target is included in the log directory name, the `+` character is replaced with an `_` character. This can lead to logs ending up in a different directory than was originally intended by the user. 

`+` is not usually too much of a problem in file names (it can be a little problematic on the CLI), but was there a specific reason for this behaviour for the log directory?

gijzelaerr:
## Summary
- Fixes the log path sanitization regex in `stimela/stimelogging.py` to preserve `+` and `@` characters, so J2000 coordinate names (e.g. `J1939+4540`) are no longer silently altered in log directory/file names
- Extracts the sanitization regex into a module-level compiled constant `_RE_LOG_SANITIZE` for clarity and testability
- Adds 7 targeted tests in `tests/test_log_sanitization.py` covering preserved characters, replaced characters, and realistic J2000 coordinate patterns

Fixes #513

## Test plan
- [x] New tests in `tests/test_log_sanitization.py` pass (7/7)
- [x] Ruff lint and format checks pass
- [ ] Existing test suite passes (CI)

### DIFF
--- a/stimela/stimelogging.py
+++ b/stimela/stimelogging.py
@@ -167,6 +167,11 @@ _boring = False
 
 LOG_DIR = "."
 
+# Characters allowed in log file paths: alphanumeric, dot, underscore, slash, at-sign, plus, hyphen.
+# The '+' and '@' characters are explicitly allowed so that e.g. J2000 coordinate names
+# (like J1939+4540) are preserved in log directory/file names (see issue #513).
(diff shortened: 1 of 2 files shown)


gijzelaerr:
Duplicate of #582 — same fix, keeping the earlier PR.

PR #596, opened 2026-06-25
Fix nested assignments being silently ignored

### LINKED ISSUE #467: Nested assignments are silently ignored.
I stumbled over this bug while working on #461. `test_scatter.yml` in the tests demonstrates the issue, but here is a simpler reproducer. I would expect the following to write the logs to `foobarbaz.txt`, but that doesn't seem to happen.

```yaml
cabs:
  sleep:
    command: sleep 1

recipe:
  assign:
    log:
      name: foobarbaz
  steps:
    sleep-1:
      cab: sleep
```

gijzelaerr:
## Summary

Fixes #467.

- The `flatten_dict` function in `update_assignments()` was defined but never called -- the line `flattened = assignments  # flatten_dict(assignments)` passed assignments through unchanged. This meant nested dict assignments like `assign: {log: {name: foobarbaz}}` were silently ignored instead of being flattened to dotted keys like `{"log.name": "foobarbaz"}`.
- Re-enables the `flatten_dict` call so nested assignment dicts are properly flattened into dotted keys before processing.
- Fixes a mutable default argument bug in the original `flatten_dict` function (`output_dict={}` shared across calls).

## Test plan

- [x] Added `test_nested_assign` test that verifies nested assign dicts for log options and recipe variables
- [x] Verified existing tests pass (test_aliasing, test_nesting, test_loop_recipe all use nested assigns and pass)
- [x] Verified flat dotted key assignments (e.g. `log.name: value`) still work

### DIFF
--- a/stimela/kitchen/recipe.py
+++ b/stimela/kitchen/recipe.py
@@ -193,11 +193,12 @@ class Recipe(Cargo):
         if not whose.assign and not whose.assign_based_on:
             return
 
-        def flatten_dict(input_dict, output_dict={}, prefix=""):
+        def flatten_dict(input_dict, prefix=""):
+            output_dict = {}
             for name, value in input_dict.items():
                 name = f"{prefix}{name}"
                 if isinstance(value, (dict, OrderedDict, DictConfig)):
-                    flatten_dict(value, output_dict=output_dict, prefix=f"{name}.")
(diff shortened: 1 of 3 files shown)


gijzelaerr:
Duplicate of #600 — same fix, keeping the batch agent's PR.

PR #601, opened 2026-06-25
Reject unknown CLI parameters (fixes #530)

### LINKED ISSUE #530: Silent acceptance of non-existent parameters
This has caught me off-guard several times whilst overhauling some of the inputs in breifast. Stimela will happily accept parameters which don't exist and will not inform the user of that fact. This can easily break/change the behaviour of recipes when a user is updating input names etc. Consider the following simple example (`recipe.yml`):
```yml
cabs:
  echo:
    command: echo
    inputs:
      msg:
        dtype: str
        required: true
        policies:
          positional: true

recipe:
  steps:
    echo-1:
      cab: echo
      params:
        msg: Hello world!
```
This can be invoked with `stimela run recipe.yml foo=bar` without issue, despite the fact that foo is neither an input, nor a field in an assign. I guess I am unsure of what the intended behaviour is in this case, but I do think that this worth thinking about.

gijzelaerr:
## Summary
- When users pass parameters on the command line that don't match any known recipe input/output, step name, config setting, or log option, stimela now raises an `AssignmentError` instead of silently ignoring them.
- This prevents the dangerous scenario described in #530 where typos in parameter names lead to wrong results with no warning.
- The check only applies to CLI override assignments. Recipe-level `assign:` blocks can still set arbitrary recipe namespace variables (existing intentional behavior).

## Changes
- `stimela/kitchen/recipe.py`: Added `elif override:` clause in `Recipe.assign_value()` that raises `AssignmentError` for unrecognized parameter keys passed from the CLI.
- `tests/test_issue530.yml`: New minimal test recipe for the issue.
- `tests/test_recipe.py`: New `test_issue530_unknown_cli_params` test covering unknown recipe-level params, unknown step-level params, nonexistent step names, and valid parameter usage.

## Test plan
- [x] Unknown recipe-level parameter (`nonexistent_param=bad`) raises error
- [x] Unknown step-level parameter (`echo-1.nonexistent_param=bad`) raises error  
- [x] Nonexistent step name (`nonexistent_step.param=bad`) raises error
- [x] Valid parameters still work correctly
- [x] Existing tests pass (aliasing, nesting, issue527, param_file, callables, conditional_skips)
- [x] Recipe-level `assign:` blocks with arbitrary variables still work (not affected by fix)

Closes #530

### DIFF
--- a/stimela/kitchen/recipe.py
+++ b/stimela/kitchen/recipe.py
@@ -356,6 +356,11 @@ class Recipe(Cargo):
             whose.update_log_options(**{subkey: value})
             if whose is self and subst is not None and "recipe" in subst:
                 subst.recipe.log[subkey] = value
+        elif override:
+            # In override mode (CLI parameters), unknown keys are an error.
+            # In non-override mode (recipe assign blocks), unknown keys are
+            # arbitrary recipe-level variables already handled by _do_assignments.
+            raise AssignmentError(f"{self.fqname}: '{key}' does not refer to a known parameter, step, or setting")
         # in override mode, assign to assign dict for future processing
         if override:
             if value is not UNSET:
--- a/tests/test_issue530.yml
+++ b/tests/test_issue530.yml
@@ -1,0 +1,21 @@
+cabs:
+  echo:
+    command: echo
+    inputs:
+      msg:
+        dtype: str
+        required: true
+        policies:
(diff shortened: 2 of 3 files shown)


gijzelaerr:
Duplicate of #585 — same fix, keeping the earlier PR.

Issue #565, opened 2026-06-24
support (filename.py)function syntax for dynamic schemas and python flavour cabs


o-smirnov:
Recent test woes with @JSKenyon led us to desire this feature. Currently dynamic schemas and python flavour cabs expect the callable to be specified as ``module.function``, and wants ``module`` to be importable.

This creates a problem with tests, but also overall it would be desirable to be able to point directly to a python file as an alternative. Suggest that we recognize a syntax (somewhat consistent with ``_include``) of ``(path/filename.py)/function`` for these cases.

ChatGPT suggests:

```python
import importlib.util

module_name = "mymodule"
file_path = "/path/to/mymodule.py"

spec = importlib.util.spec_from_file_location(module_name, file_path)
module = importlib.util.module_from_spec(spec)
spec.loader.exec_module(module)

module.some_function()
```

Also, consistent with https://stimela.readthedocs.io/en/latest/fundamentals/include.html#include-paths-and-dangers-thereof, a path starting with ``./`` needs to be interpreted relative to the location of the yml file where it appears. Other paths can be interpreted relative to the CWD.

PR #603, opened 2026-06-29
Move to src layout and fix tests

JSKenyon:
There are several issues which are caused by having a flat project layout - see https://packaging.python.org/en/latest/discussions/src-layout-vs-flat-layout/. This PR moves stimela to a src layout, ensuring that an install no longer packages the tests. While this can also be accomplished by excluding the tests in the `pyproject.toml` file, that approach is less comprehensive as editable installs will often point to the source directory. In the flat case, this means that the test directory may remain accessible (and thus able to interfere with other projects which have an identically named test directory).

Making this change broke some of the tests as there is no longer an installed `tests` module to import from. I believe that I have fixed all of the problems. To do so I had to move the callables and dynamic schema examples into the main package under `src/stimela/examples`. This should be harmless and may ultimately be avoidable once #565 is properly supported and battle tested; I didn't want to couple this PR to that.

Note that this change should likely be applied to scabha too, for similar reasons.


o-smirnov (approved):
LGTM!

